PR #3, opened 2017-10-12
ra_log_wal: Ignore exception if the ra node is gone

dumbbell:
In `complete_batch()`, we want to notify the ra node that its data was written. If the ra node is a registered process, sending it a message will raise an exception if the process is gone. As we can't do much about this, let's ignore the exception.


kjnilsson:
This is being address as part of the `ra_log_file` hardening work.

PR #13, opened 2018-01-31
Use ra.app.src file which rebar uses to autogenerate modules

liveforeverx:
No need to track and manually add modules to ra.app file.

`_build` is a default build location for `rebar3`


kjnilsson:
Thank you. There is no manual step as erlang.mk generates the .app file at build time. the rebar.config is only there to allow `ra` to be used as a dependency by rebar3, for development and testing erlang.mk should be used, at least for now.

PR #14, opened 2018-02-07
Partitions test

hairyhum:
partition_SUITE - publish/consume with restarting nodes and single node being blocked out.

jepsen_like_partition_SUITE - publish consume from a node in the cluster, counting sent,
confirmed, unconfirmed, delivered and acked messages. Uses nemesis process to create random
partitions.

### DIFF
--- a/test/jepsen_like_partitions_SUITE.erl
+++ b/test/jepsen_like_partitions_SUITE.erl
@@ -5,8 +5,8 @@
 
 all() -> [
  publish_ack_inet_tcp_proxy
- ,
- publish_ack_iptables
+ % ,
+% publish_ack_iptables
 ].
 
 init_per_testcase(TestCase, Config0) ->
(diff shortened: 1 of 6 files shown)


hairyhum:
What's curious, is that partitions test has redeliveries in blocking test case. Not sure if the test case is usable, because nodes are blocked with inet_tcp_proxy, but there are no reelections, the message is settled, but still redelivery happens.

test/jepsen_like_partitions_SUITE.erl
```diff
@@ -0,0 +1,357 @@
...
+get_random_node(Nodes) ->
+    lists:nth(rand:uniform(length(Nodes)), Nodes).
+
+get_random_node(Exceptions, Nodes) ->
+    PossibleNodes = Nodes -- Exceptions,
+    get_random_node(PossibleNodes).
+
+
+data_dir() ->
+    {ok, Cwd} = file:get_cwd(),
```
kjnilsson: use the ct `priv_dir` here.

hairyhum:
Currently there are some acks being lost both for publishers and consumers. Can reproduce with both iptables and inet_tcp_proxy.
Iptables test should be run as sudo.

PR #19, opened 2018-08-03
Rename committed_query to local_query.

hairyhum:
The committed_query function name is confusing, because it makes
you think the query is being committed.
Since it's already marked as dirty in the internal protocol, why
not call it like this?
This also highlights a difference between consistent and dirty.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -50,8 +50,8 @@
 
 4. Perform a query
 
-    As we know the previous command achieved consensus a committed query to the leader is fine.
+    As we know the previous command achieved consensus a local query to the leader is fine.
 
     ```
-    {ok, {IdxTerm, 5}, Leader} = ra:committed_query({node1, node()}, fun (S) -> S end),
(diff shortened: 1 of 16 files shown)


kjnilsson:
I think the only alternative here is to name it `query_committed` as it _is_ querying committed data, not dirty (i.e. in flight) changes. The internal message should be updated accordingly.

hairyhum:
It's already called dirty in the internal API. 
I understand that the state machine state cannot be not committed, since commands are applied after commit, so how can it be "in flight"?

kjnilsson:
It's not possible to query in flight entries as it stands (it would be theoretically possible to implement a dirty query if needed but I don't really see a use case) hence the term dirty query needs to be removed completely. I suggest we rename this to `query_committed` both externally and internally. Perhaps we should should then also match the consistent_query and rename it to `query_consistent`

kjnilsson:
The use of "dirty" internally was a mistake on my part.

PR #20, opened 2018-08-03
Add timeout to API functions, which use default timeout.

hairyhum:
Default timeout of 5 seconds may be not enough in some cases.
The API functions which can have a configurable timeout should have it.
Removed default timeout from ra_node_proc API. Internal APIs are
easier to reason about when they don't use constants.


hairyhum:
Regenerated docs

kjnilsson:
Yes good point. Thanks.

PR #25, opened 2018-10-30
Pluggable snapshotting

hairyhum:
Make snapshot module pluggable.

Add the `ra_snapshot` behaviour module.

Change snapshot API to separate metadata (Index, Term and Servers) from the snapshot state.

Snapshot lifetime:

- Machine issues `{release_cursor, Index, State}`
- If log module decides to take a snapshots it calls the `prepare` callback
- `prepare` callback returns a reference to the data to be saved (can be the machine state or a mnesia checkpoint)
- ra_snapshot_writer calls the `write` callback with the reference and metadata to save it to disk
- when a leader sends a snapshot it reads the metadata and the opaque snapshot data using the `read` callback
- when a follower receives the data it saves it to disk using the `save` callback
- to create a machine state after `save` it calls `install`, which will restore the machine state from the data saved
- when a node starts, it calls `recover` to read existing snapshot and restore the machine state.


src/ra_server.erl
```diff
@@ -729,6 +733,10 @@ handle_follower({ra_log_event, {written, _} = Evt},
                 State00 = #{log := Log0}) ->
     State0 = State00#{log => ra_log:handle_event(Evt, Log0)},
     evaluate_commit_index_follower(State0);
+handle_follower({ra_log_event, {snapshot_written, _, _, _} = Evt},
```
kjnilsson: Is this clause needed?

src/ra_log_snapshot.erl
```diff
@@ -67,6 +77,12 @@ read(File) ->
             Err
     end.
 
+-spec install(term(), file:filename()) -> {ok, term()}.
+install(Data, _File) -> {ok, Data}.
+
+-spec recover(file:filename()) -> {ok, meta(), term()} | {error, invalid_format | {invalid_version, integer()} | checksum_error | file_err()}.
```
kjnilsson: This line is tooooooooooooooooooooooooooooooooooo long.

src/ra.erl
```diff
@@ -233,13 +244,20 @@ start_cluster(ClusterName, Machine, ServerIds) ->
                    ra_server:machine_conf(), [ra_server_id()]) ->
     ok | {error, term()}.
 start_server(ClusterName, ServerId, Machine, ServerIds) ->
+    start_server(ClusterName, ServerId, Machine, ?DEFAULT_SNAPSHOT_MODULE, ServerIds).
+
+-spec start_server(ra_cluster_name(), ra_server_id(),
+                   ra_server:machine_conf(), module(), [ra_server_id()]) ->
+    ok | {error, term()}.
+start_server(ClusterName, ServerId, Machine, SnapshotModule, ServerIds) ->
```
kjnilsson: As discussed out of bound the snapshot module should be provided as an optional callback from the `ra_machine` implementation rather than configuration. `ra_machine` and `ra_snapshot` are typically linked together and being able to configure them independently isn't ideal.

src/ra_log.erl
```diff
@@ -77,6 +77,7 @@
          snapshot_state :: maybe({ra_index(), ra_term(),
                                   maybe(file:filename())}),
          snapshot_interval = ?SNAPSHOT_INTERVAL :: non_neg_integer(),
+         snapshot_module = ?DEFAULT_SNAPSHOT_MODULE :: module(),
```
kjnilsson: no default should be needed.

src/ra_log.erl
```diff
@@ -139,7 +141,7 @@ init(#{uid := UId} = Conf) ->
             [File | _] ->
                 %% TODO provide function that only reads the index and term
                 %% of the snapshot file.
-                {ok, {SI, ST}} = ra_log_snapshot:read_indexterm(File),
+                {ok, {SI, ST}} = SnapModule:read_indexterm(File),
```
kjnilsson: module calls like this should go through the behaviour they are implemented: `ra_snapshot(SnapModule, File)` - This helps validating these calls can keep the dynamic dispatch separated from "normal" code

src/ra_log.erl
```diff
@@ -467,37 +470,59 @@ fetch_term(Idx, #state{cache = Cache, uid = UId} = State0) ->
...
+-spec recover_snapshot(State :: ra_log()) ->
+    maybe({ok, ra_snapshot:meta(), term()}).
+recover_snapshot(#state{snapshot_state = undefined}) ->
+    undefined;
+recover_snapshot(#state{snapshot_state = {_, _, File},
+                        snapshot_module = SnapModule}) ->
+    case SnapModule:recover(File) of
+        {ok, Meta, Data} ->
+            {ok, Meta, Data};
+        {error, enoent} ->
```
kjnilsson: should we handle this? when would we have a snapshot state but no file on disk?

src/ra_log_snapshot.erl
```diff
@@ -1,9 +1,15 @@
...
 
+-behaviour(ra_snapshot).
+
 -export([
-         write/2,
+         prepare/2,
+         write/3,
+         save/3,
          read/1,
+         install/2,
```
kjnilsson: As discussed I'd like `install/2` removed in favour of just using `recover/1` - this will help when we want to implement chunked transfer that doesn't maintain the full state in memory.

src/ra_server.erl
```diff
@@ -411,9 +411,13 @@ handle_leader({ra_log_event, {written, _} = Evt}, State0 = #{log := Log0}) ->
     {State, Rpcs} = make_pipelined_rpcs(State1),
     {leader, State, [{send_rpcs, Rpcs},
                      {incr_metrics, ra_metrics, [{3, Applied}]} | Effects]};
+handle_leader({ra_log_event, {snapshot_written, _, _, _} = Evt},
```
kjnilsson: do we need this?

src/ra_snapshot.erl
```diff
@@ -0,0 +1,44 @@
...
+%% Turn the current state into immutable reference.
+-callback prepare(Index :: ra_index(), State :: term()) -> Ref :: term().
+
+%% Saves snapshot from external state to disk.
+%% Runs in a separate process.
+%% External storage should be available to read
+-callback write(Location :: file:filename(),
+                Meta :: meta(),
+                Ref :: term()) ->
+    ok | {error, file_err()}.
```
kjnilsson: is file_err the only errors we can imagine here?

src/ra_snapshot.erl
```diff
@@ -0,0 +1,44 @@
...
+    {ok, Meta :: meta(), Data :: term()} |
+    {error, invalid_format | {invalid_version, integer()} | checksum_error | file_err()}.
+
+%% Dump the snapshot data to disk withtou touching the external state
+-callback save(Location :: file:filename(), Meta :: meta(), Data :: term()) ->
+    ok | {error, file_err()}.
+
+%% Side-effect function
+%% Deserialize the snapshot to external state.
+-callback install(Data :: term(), Location :: file:filename()) ->
```
kjnilsson: As already mentioned I think this should be removed in favour of `recover/1`

src/ra_snapshot.erl
```diff
@@ -0,0 +1,44 @@
...
+    {ok, State :: term()} | {error, term()}.
+
+%% Side-effect function
+%% Recover machine state from file
+-callback recover(Location :: file:filename()) ->
+    {ok, Meta :: meta(), State :: term()} | {error, term()}.
+
+%% Only read index and term from snapshot
+-callback read_indexterm(Location :: file:filename()) ->
```
kjnilsson: maybe this should just be `read_meta/1` now? It's likely that meta data will be separately written  just using term_to_binary of the meta data triple, maybe even for the default snapshotter.

src/ra_log.erl
hairyhum: When the node is starting, recover expects undefined. This behaviour was there before, I just moved it to this function.

kjnilsson:
commits are included in #30

PR #30, opened 2018-11-22
Snapshot improvements

kjnilsson:
Implements pluggable snapshot strategies (using the `ra_snapshot` behaviour) as well as chunked snapshot replication.

edocs will be updated separately to not cloud the PR too much


src/ra_snapshot.erl
```diff
@@ -0,0 +1,354 @@
...
+
+%% Read the snapshot from disk into serialised structure for transfer.
+-callback read(ChunkSizeBytes :: non_neg_integer(),
+               Location :: file:filename()) ->
+    {ok, Meta :: meta(), ChunkState,
+     [ChunkThunk :: fun((ChunkState) -> {binary(), ChunkState})]} |
+    {error, term()}.
+
+%% begin a stateful snapshot acceptance process
+-callback begin_accept(SnapDir :: file:filename(),
```
hairyhum: Can accept functions be in a separate module? I think most snapshotters just need to transfer a file of several files.

hairyhum:
I'm going to try integrating this with mnevis mnesia based snapshots.

src/ra_snapshot.erl
kjnilsson: no you need to be able to "connect both ends" the acceptor functions is the receiver of the chunks returned in read. how to interpret these chunks is snapshotter specific.

hairyhum: I think there can be some helper functions to just write files.

kjnilsson: ok what can we add on top of `file:write_file/2` ?

Thinking about it should chunks be term() rather than binary() if you want to say, wrap it with some info around filenames and so forth?

kjnilsson:
@hairyhum I'm going to merge this but not produce another hex release until you've had a chance to test it. We can still refine it and add helpers if we feel it is needed.

PR #35, opened 2018-12-04
support gen-batch-server 0.6

benoitc:
with gen-batch-server >= 0.6x cast doesn't pass the requester pid
anymore. This change fix it.


benoitc:
seems like it is requiring a new version of Make not provided by the vm.

kjnilsson:
Looks good, thanks. The only change I can see is that the `wal_op()` type in `ra_log_wal` needs to be updated.

benoitc:
@kjnilsson done

PR #37, opened 2018-12-12
Do not configure data_dir. Use the current dir by default.

hairyhum:
Do not set the default `data_dir`. Use the current dir if it's not set.

The applications can check if it's set in the config file by checking the env variable.

### DIFF
--- a/src/ra_env.erl
+++ b/src/ra_env.erl
@@ -10,13 +10,15 @@
               ]).
 
 data_dir() ->
-    case application:get_env(ra, data_dir) of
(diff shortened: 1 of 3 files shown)


src/ra_env.erl
```diff
@@ -10,13 +10,15 @@
               ]).
 
 data_dir() ->
-    case application:get_env(ra, data_dir) of
+    DataDir = case application:get_env(ra, data_dir) of
```
kjnilsson: Reformat case expression to line up `case` and `end`.

PR #39, opened 2018-12-17
Change ra_snapshot:read API to mirror ra_snapshot:accept

hairyhum:
Split ra_snapshot:read/2 to begin_read/2 and read_chunk/2
The goal of this change is to give a clearer API to snapshot
implementers, so they can think of read and accept stages with
similar functions.

### DIFF
--- a/src/ra_server_proc.erl
+++ b/src/ra_server_proc.erl
@@ -1189,16 +1189,15 @@ receive_snapshot_timeout() ->
                         ?DEFAULT_RECEIVE_SNAPSHOT_TIMEOUT).
 
 send_snapshots(Me, Id, Term, To, ChunkSize, SnapState) ->
-    {ok, Crc,
+    {ok,
      {LastIdx, LastTerm, Config},
-     ReadState} = ra_snapshot:begin_read(ChunkSize, SnapState),
+     ReadState} = ra_snapshot:begin_read(SnapState),
 
     RPC = #install_snapshot_rpc{term = Term,
                                 leader_id = Id,
                                 last_index = LastIdx,
                                 last_term = LastTerm,
-                                last_config = Config,
-                                crc = Crc},
+                                last_config = Config},
 
     Result = read_chunks_and_send_rpc(RPC, To, ReadState, 1, ChunkSize, SnapState),
     ok = gen_statem:cast(Me, {To, Result}).
--- a/src/ra_server_proc.erl
+++ b/src/ra_server_proc.erl
@@ -1209,9 +1209,9 @@ read_chunks_and_send_rpc(RPC0, To, ReadState0, Num, ChunkSize, SnapState) ->
            end,
     {ok, Data, ContState} = ra_snapshot:read_chunk(ReadState0, ChunkSize, SnapState),
     ChunkFlag = case ContState of
-        {next, _} -> next;
-        last -> last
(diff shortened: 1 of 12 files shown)


hairyhum:
With ChunkSize being an argument to read_chunk, we can change the chunk size in flight. I'm not sure how valuable it is, but it's possible now.

hairyhum:
I'm not sure why do we need Crc to be a part of the protocol and restrict it's type. `complete_accept` and `validate` functions should validate the data integrity of the snapshot and for `ra_log_snapshot` the Crc is a part of on-disk data. Do we need to transfer it separately? If we do, is it a part of the snapshot metadata?

kjnilsson:
For multi file transfers a single Crc doesn't make sense so we can probably remove this and include a crc as part of the chunk term

src/ra_server_proc.erl
```diff
@@ -1186,31 +1186,35 @@ receive_snapshot_timeout() ->
...
+    Result = read_chunks_and_send_rpc(RPC, To, ReadState, 1, ChunkSize, SnapState),
     ok = gen_statem:cast(Me, {To, Result}).
+
+read_chunks_and_send_rpc(RPC0, To, ReadState0, Num, ChunkSize, SnapState) ->
+    Conf = case Num of
+               1 -> RPC0#install_snapshot_rpc.last_config;
+               _ -> undefined
+           end,
+    {ok, Data, ContState} = ra_snapshot:read_chunk(ReadState0, ChunkSize, SnapState),
+    ChunkFlag = case ContState of
```
kjnilsson: formatting

src/ra_snapshot.erl
```diff
@@ -310,16 +317,25 @@ delete(Dir, {Idx, Term}) ->
...
     Location = make_snapshot_dir(Dir, Idx, Term),
-    Mod:read(ChunkSizeBytes, Location).
+    Mod:begin_read(ChunkSizeBytes, Location).
+
+
+-spec read_chunk(ReadState, ChunkSizeBytes :: non_neg_integer(), State :: state()) ->
+    {ok, Data :: term(), {next, ReadState} | last}  |
+    {error, term()} when ReadState :: term().
+read_chunk(ReadState, ChunkSizeBytes, #?MODULE{module = Mod,
+                                directory = Dir,
```
kjnilsson: align fields

src/ra_snapshot.erl
```diff
@@ -310,16 +317,25 @@ delete(Dir, {Idx, Term}) ->
...
-
--spec read(ChunkSizeBytes :: non_neg_integer(), State :: state()) ->
-    {ok, Crc :: non_neg_integer(), Meta :: meta(), ChunkState,
-     [ChunkThunk :: fun((ChunkState) -> {binary(), ChunkState})]} |
-    {error, term()} when ChunkState :: term().
-read(ChunkSizeBytes, #?MODULE{module = Mod,
+-spec begin_read(ChunkSizeBytes :: non_neg_integer(), State :: state()) ->
+    {ok, Crc :: non_neg_integer(), Meta :: meta(), ReadState} |
+    {error, term()} when ReadState :: term().
+begin_read(ChunkSizeBytes, #?MODULE{module = Mod,
```
kjnilsson: align fields below

PR #42, opened 2018-12-20
Ra machine apply 3

### LINKED ISSUE #38: RFC: remove list of effects from ra_machine:apply/4
`ra_machine:apply/4` takes a list of effects as it's 3rd argument. This is done for efficiency so that all effects in a batch of commands are cons-ed to the list, avoiding any list concatenation. The list of effects is then reversed at the end of the apply batch and executed.

This does mean that implementors of `ra_machine` need to return the effects for a single command in the reverse order they want them actioned. I think this isn't a great API and would like to remove the effects from the `apply/4` function making it `apply/3` and accept the effects from in the order they should be implemented. I think we can do this efficiently by simply gather effects in a nested list rather than a flat list. this would also machines to return single effects without wrapping them into a list.

I also suggest optionally moving the reply to the second slot in the return tuple allowing implementors to leave out the effects all together.

Current signature:

```
-callback 'apply'(command_meta_data(), command(), effects(), State) ->
    {State, effects(), reply()}.
```

Proposed signature:

```
-callback 'apply'(command_meta_data(), command(), State) ->
    {State, reply(), effect() | [effect()]} | {State, reply()}.
```

WDYT @hairyhum @michaelklishin @dcorbacho @benoitc @acogoluegnes

kjnilsson:
Change ra_machine apply/4 into apply/3

    This simplifies the ra_machine apply behaviour function not to take a
    list of effects and necessitate that any further effects are cons-ed to
    the existing list. Instead implemetors just return a list of effects in
    execution order.

    Also extended and amended the return type to avoid creating unnecessary
    lists and tuple slots.

Closes #38.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-# Ra: a Raft Implementation for Erlang and Elixir
 
 [![Build Status](https://travis-ci.org/rabbitmq/ra.svg?branch=master)](https://travis-ci.org/rabbitmq/ra)
 
--- a/docs/edoc-info
+++ b/docs/edoc-info
@@ -1,3 +1,3 @@
 %% encoding: UTF-8
 {application,ra}.
-{modules,[ra,ra_directory,ra_env,ra_machine]}.
+{modules,[ra,ra_directory,ra_env,ra_machine,ra_server,ra_snapshot]}.
--- a/docs/internals/INTERNALS.md
+++ b/docs/internals/INTERNALS.md
@@ -43,8 +43,8 @@ implemented:
 ```erlang
 -callback init(Conf :: machine_init_args()) -> state().
 
--callback 'apply'(command_meta_data(), command(), effects(), State) ->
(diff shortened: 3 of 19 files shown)


kjnilsson:
I ran some local throughput tests through RabbitMQ where the rabbit_fifo statemachine emits a lot of effects and there was no noticeable difference. We are lacking more granular benchmarking tests in ra though.

I think that TODO is unlikely to have a huge negative effect as we don't really convert that many effects into gen_statem actions. I'm happy for that to remain for a rainy day.

michaelklishin:
Should we do a new release or switch ra-examples to `master`? What's the plan for updating Ra callers in RabbitMQ?

kjnilsson:
There is a rabbitmq-server PR linked that updates the state machine to the new api

kjnilsson:
examples can be updated after we do the next hex release.

PR #45, opened 2018-12-31
Reject user commands

dcorbacho:
Ra machines have the option now to reject commands before appending to the log or during the `apply`. Machines can implement the optional callbacks `filter` and `init_filter`.

## Types of Changes

## Checklist


kjnilsson:
This is a good start.

I think the reply mechanism for filtered commands should be based on the command reply mode rather than on the state machine. The state machine should only provide the reason for the command being filtered and an appropriate response or notification should be generated.

If the command reply mode is `after_log_append` or `await_consensus` an appropriate reply should be returned to the caller.

If the command reply mode is `notify_on_consensus` a `ra_event` should be generated like is done when a pipelined command is sent to a follower by mistake.

I suggest the following format:

`{ra_event, {rejected, {filtered, FilteredReason, Correlation}}}`

This PR also needs tests for this functionality. I suggest adding them to `ra_machine_int_SUITE.erl`

michaelklishin:
@dcorbacho @kjnilsson should this also be closed together with https://github.com/rabbitmq/rabbitmq-server/pull/1813?

PR #46, opened 2019-01-03
Optimisation

kjnilsson:
Various optimisation and minor fixes found during throughput testing using the noop machine.

Optimised hot code paths such as AppendEntriesRpc generation and ra log cache truncation

Also ensure that pipelining of very behind followers isn't unnecessarily slow.

Fix rare crash bug when an empty segment is created and then the server is restarted.

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -266,10 +266,6 @@ recover(#{id := Id,
     Log = ra_log:release_resources(1, Log0),
     State#{log => Log}.
 
-% the peer id in the append_entries_reply message is an artifact of
-% the "fake" rpc call in ra_proxy as when using reply the unique reference
-% is joined with the msg itself. In this instance it is treated as an info
-% message.
 -spec handle_leader(ra_msg(), ra_server_state()) ->
     {ra_state(), ra_server_state(), ra_effects()}.
 handle_leader({PeerId, #append_entries_reply{term = Term, success = true,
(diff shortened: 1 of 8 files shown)


src/ra_server.erl
```diff
@@ -1380,25 +1421,25 @@ peers_not_sending_snapshots(State) ->
...
-                        NI - MI < ?MAX_PIPELINE_DISTANCE;
-                    (_, #{commit_index_sent := CI,
-                          next_index := NI,
-                          match_index := MI}) when CI < CommitIndex ->
-                        % the commit index has been updated
-                        NI - MI < ?MAX_PIPELINE_DISTANCE;
-                    (_, _) ->
-                        false
-                end, peers(State)).
+% pipelineable_peers(#{commit_index := CommitIndex,
```
michaelklishin: Is this dead code necessary?

test/ra_SUITE.erl
```diff
@@ -478,6 +478,7 @@ snapshot_installation_with_call_crash(Config) ->
               fun () ->
                       {ok, {N1Idx, _}, _} = ra:local_query({N1, node()},
                                                            fun ra_lib:id/1),
+                      ct:pal("N1 idx ~w", [N1Idx]),
```
michaelklishin: Should this be removed?

michaelklishin:
The noop machine goes up to 50K log entries processed per second on this branch.

Issue #10, opened 2018-01-04
ra:consistent_query/2 should not go through the log


kjnilsson:
As the query gets persisted in the log we need it to be an MFA rather than a fun.

PR #62, opened 2019-03-12
Consistent query without going through the log

hairyhum:
Add new RPCs: read_only_heartbeat/read_only_heartbeat_reply.

When a leader gets the query, it will save the current commit index,
issue a heartbeat message and wait for replies.

After a quorum of replies, it executes the query as soon as respective
commit index applied.

TODO:
- verification
- no-op message if the leader is new?
- waiting_apply_index data structure

Fixes: https://github.com/rabbitmq/ra/issues/10


hairyhum:
Read only heartbeat works similarly to append_entries_rpc, updating term. But it does not update the current index.

hairyhum:
Step 1 in section 6.4 of the thesis mentions a no-op entry if the leader was just elected. I'm not sure that ra needs that, because it has the pre-vote stage.
@kjnilsson can you please clarify?

kjnilsson:
@hairyhum Ra commits a noop entry at the beginning of each term. There is a flag called `cluster_change_permitted` that indicates whether the noop has been committed or not.

kjnilsson:
I think the general implementation is fine but I'd like it to replace the `ra:consistent_query` API - effectively completing https://github.com/rabbitmq/ra/issues/10

hairyhum:
Does it mean that I need to check `cluster_change_permitted` before issuing a heartbeat? For the most of it I've copied `append_entries_rpc` behaviour

hairyhum:
@kjnilsson ready for another review. Can you also clarify about `cluster_change_permitted`?

michaelklishin:
@hairyhum can you please resolve conflicts with master?

kjnilsson:
Generally the approach looks fine. We'll need some unit tests in `ra_server_SUITE` for the new protocol messages. There is also some formatting that looks a bit wonky (probably after rename excercise).

src/ra_server.erl
```diff
@@ -1901,6 +1994,102 @@ drop_existing({Log0, [{Idx, Trm, _} | Tail] = Entries}) ->
...
+heartbeat_rpc_quorum(Ref, PeerId, #{waiting_heartbeats := WH0,
+                                    cluster := Cluster} = State) ->
+    case {maps:get(Ref, WH0, none), maps:get(PeerId, Cluster, none)} of
+        %% Old message. Do nothing
+        {none, _} -> {false, State, []};
+        %% Peer is not in the current peers list
+        {_, none} -> {false, State, []};
+        {WaitingPeers0, _} ->
+            PeerIds = peer_ids(State),
+            MissingPeerIds = lists:filter(fun(PeerIdInCluster) ->
```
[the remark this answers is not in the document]
hairyhum: Yes. Should there be a different approach?
BTW, this should actually be checked later in the final clause of this function.

hairyhum:
Does it mean that we should delay queries until `cluster_change_permitted` is true?
Also it does not look like the code handles leader->follower transition. Need to fix that. Is there some sort of transition callback I can hook in?
As I see that, when the leader becomes follower - it should clean up all the temp state for queries and send the query to the new leader.

src/ra_server.erl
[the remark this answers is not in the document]
kjnilsson: No I think this is fine. Perhaps just pop a little comment in there to explain for slow folks like me. :)

kjnilsson:
Yes you `cluster_change_permitted` should be true before you serve any queries from a leader so it's another thing to wait for. if you hit the scenario where it isn't true you don't really need to send heartbeats but instead just delay running queries until it is true. (which should be very soon as it is set to true as soon as the noon command is committed after a leader change)

kjnilsson:
When the leader steps down another approach would be to somehow return a redirect response so that the query will be re-run. Like the redirect that is done for leader_calls. That would save having a special mechanism for that case.

kjnilsson:
This redirect code here could possibly be used? https://github.com/rabbitmq/ra/blob/3ea6daa2ab8d6fb237d5d09f9a3082fcf8bbd3c4/src/ra_server_proc.erl#L1201-L1213

hairyhum:
I'm less worried about redirect and more about hooking into the right place when transition happen. Should it be done in `state_enter`?

hairyhum:
I can see that `cluster_change_permitted` may change back and forth, which makes me woner if it should be checked on all callbacks or just the initial request. Technically we are checking that the node is the leader for `read_index`.
Do we have some sort of a formal model for this thing? Because getting it from the code is hard.

kjnilsson:
I'm not sure what you're after. This flag changes to true once the noop command has been committed after which is stays true for the duration of this term. A leader shouldn't serve consistent read requests unless this is true.

hairyhum:
Thanks for clarification, I didn't know that. That's why I was asking for a formal model.
So it should be enough to check that this flag is true when receiving a `consistent_query` call request?
As for leader stepping down, is this the place it should be handled in: https://github.com/rabbitmq/ra/blob/master/src/ra_server_proc.erl#L391 ?

kjnilsson:
I think as long as the flag is set by the time you want to run the query it should be ok.

Yes that is probably the only place you can hook into that particular state change. I definitely would consider just triggering redirects rather than adding a new mechanism if you can.

hairyhum:
The `rabbit_server` code needs refactoring. It's really hard to navigate that.

hairyhum:
Ready for another review

include/ra.hrl
```diff
@@ -134,6 +136,18 @@
          last_index :: ra_index(),
          last_term :: ra_term()}).
 
+-record(heartbeat_rpc, {
```
kjnilsson: keep the same record formatting as the other records

src/ra_server.erl
```diff
@@ -1901,6 +2015,120 @@ drop_existing({Log0, [{Idx, Trm, _} | Tail] = Entries}) ->
...
+            WH1 = WH0#{ Ref => WaitingPeers},
+            {State#{waiting_heartbeats => WH1}, Effects}
+    end.
+
+heartbeat_rpc_effects(PeerIds, Ref, #{current_term := Term, id := Id}) ->
+    lists:map(fun(PeerId) ->
+        {send_rpc, PeerId, #heartbeat_rpc{ref = Ref,
+                                                term = Term,
+                                                leader_id = Id}}
+    end,
```
kjnilsson: formatting

src/ra_server_proc.erl
```diff
@@ -385,11 +398,13 @@ leader({call, From}, {log_fold, Fun, Term}, State) ->
     fold_log(From, Fun, Term, State);
 leader(EventType, Msg, State0) ->
     case handle_leader(Msg, State0) of
-        {leader, State1, Effects} ->
-            {State, Actions} = ?HANDLE_EFFECTS(Effects, EventType, State1),
+        {leader, State1, Effects1} ->
```
kjnilsson: why the rename?

src/ra_server.erl
```diff
@@ -589,6 +649,22 @@ handle_candidate(#append_entries_rpc{leader_id = LeaderId},
     % term must be older return success=false
     Reply = append_entries_reply(CurTerm, false, State),
     {candidate, State, [{cast, LeaderId, {id(State), Reply}}]};
+handle_candidate(#heartbeat_rpc{ term = Term } = Msg,
```
kjnilsson: no need to put spaces between `{` and the field name

src/ra_server_proc.erl
hairyhum: To have the same number in state and events on each step. It can be hard to track numbers separately.

hairyhum:
The problem with attaching this logic to `append_entries_rpc` is that it extends `append_entries_rpc` too much. It's already not easy to reason about it.
Also there may be no `append_entries_rpc` requests, but queries should execute. And heartbeats should be confirmed with the heartbeat ref, not the index. 
I don't see how that would work in top of `append_entries_rpc` without adding additional complexity. If you can - please explain that in more detail.

hairyhum:
We may generalise rpcs somehow, wrapping them in `{rpc, Term, LeaderId, RPC}` and `{reply, Term, Reply}` form and handle term comparison separately.

src/ra_server_proc.erl
kjnilsson: Generally (but not consistently) the last assignment should be number less. So the first incarnation use the 0 suffix which is incremented until the very last incarnation which has no number. It's just an unofficial pattern that emerged but it makes it easy to see which one is the final state version.

kjnilsson:
What I had in mind was:

Add a single `ref` field to `append_entries_rpc` and the reply that is always returned by the followers (unless it is undefined). If it is set the follower will reply immediately (echo). This way the consistency checking is only done in one place which should be safer. When a leader receives a reply with a ref it will evaluate it accordingly. Sending rpcs to all followers is easy using the `ra_server:send_all_rpcs/1` function as is done here: https://github.com/rabbitmq/ra/blob/260f0afe244476d999bb9f035994a79a8350ad71/src/ra_server.erl#L563-L568

hairyhum:
@kjnilsson are you ok with the current state of this PR or you think it has to use `append_entries_rpc`?

hairyhum:
@kjnilsson this needs more work. Consistent query fails to detect followers coming back. I will need your help in navigating the code.

kjnilsson:
Suggested design that augments the `append_entries_rpc` message to re-use all the liveness handling and fixes that have accumulated over time: https://gist.github.com/kjnilsson/ff0d79afc88bb583c95a9b2a64a31705

hairyhum:
Updated the branch to use read_index instead of ref for heartbeats. This may be used when moving to append_entries_rpc.

hairyhum:
Removed waiting for apply index, because apply index should always be up-to-date.

Refactored names of variables.

Made waiting for heartbeats into a queue.

src/ra_server.erl
```diff
@@ -67,7 +68,10 @@
       aux_state => term(),
       condition => ra_await_condition_fun(),
       condition_timeout_effects => [ra_effect()],
-      pre_vote_token => reference()
+      pre_vote_token => reference(),
+      query_index => non_neg_integer(),
```
kjnilsson: Are these new fields optional or should they use `:=` - i.e. be required to be in the state map

src/ra_server_proc.erl
```diff
@@ -325,6 +328,16 @@ leader({call, From}, {state_query, Spec},
        #state{server_state = ServerState} = State) ->
     Reply = {ok, do_state_query(Spec, ServerState), id(State)},
     {keep_state, State, [{reply, From, Reply}]};
+leader({call, From}, {consistent_query, QueryFun},
+       #state{server_state = ServerState0} = State0) ->
+    %% TODO: make sure there is a command in the current term
```
kjnilsson: is this TODO still valid?

hairyhum:
@kjnilsson ready for another round of review.

PR #90, opened 2019-04-29
Segment writer simplification

kjnilsson:
Refactor segment writer
So that it no longer keeps files open as this is a likely pre-mature
optimisation that only increases complexity. The segment writer will now
also always keep at least one segment to ensure segment names are never
re-used.


hairyhum:
I don't think I understand how this thing works and able to review that on my own.

PR #98, opened 2019-06-20
Metrics tweaks and meta data optimisations

kjnilsson:
Allow the metrics key to be specified through config.

Also a commit to reduce cpu use of the meta data store in low/medium throughput scenarios.


gerhard:
When put together, this is the current state of this:

image

michaelklishin:
Should this be merged then? ;)

michaelklishin:
:shipit:

gerhard:
That's the first thing that we intend to do today, with a fresh pair of eyes. 👍 👀

PR #104, opened 2019-08-01
Retry on server data directory deletion

acogoluegnes:
Retry attempts take place in a different process. ETS table cleaning is
done even if the data directory deletion fails.

There were some spurious failures on CI, this commit tries to fix this.

### DIFF
--- a/src/ra_server_sup_sup.erl
+++ b/src/ra_server_sup_sup.erl
@@ -104,7 +104,7 @@ delete_server(NodeId) ->
     end.
 
 delete_server_rpc(RaName) ->
-    ?INFO("Deleting server ~w and it's data directory.~n",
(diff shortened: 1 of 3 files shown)


src/ra_server_sup_sup.erl
```diff
@@ -104,32 +104,59 @@ delete_server(NodeId) ->
...
+            spawn(fun() ->
+                      retry(DeleteFunction, 2)
+                  end)
+    end.
+
+retry(_Func, 0) ->
+    exhausted;
+retry(Func, Attempt) ->
+    % do not retry immediately
+    timer:sleep(100),
```
kjnilsson: I'd set a larger default sleep here I think, say 5s. Also maybe move the retry function into `ra_util`

Issue #112, opened 2019-08-20
Implement leadership transfer extension


kjnilsson:
Raft thesis §3.10

This would be useful for balancing leaders across the erlang cluster.

Api suggestion:

`ra:transfer_leadership(SomeServerId, TargetServerId)`

When the leader receives this call it will:

 * send a message to the target server to immediately start an election (possibly skip straight to candidate state, bypassing pre-vote).

* enter an awaiting state where it will respond to vote requests but not append any more entries to it's log.

* once it knows about a new leader it will enter follower state as normal

* if it times out it will re-enter leader state

PR #116, opened 2019-09-05
Leadership transfer extension

dcorbacho:
Closes https://github.com/rabbitmq/ra/issues/112

## Types of Changes

## Checklist

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -67,7 +67,8 @@
       effective_machine_module := module(),
       aux_state => term(),
       condition => ra_await_condition_fun(),
-      condition_timeout_effects => [ra_effect()],
+      condition_timeout_changes => #{transition_to := ra_state(),
+                                     effects := [ra_effect()]},
       pre_vote_token => reference(),
       query_index := non_neg_integer(),
       queries_waiting_heartbeats := queue:queue({non_neg_integer(), consistent_query_ref()}),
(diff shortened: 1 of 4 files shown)


src/ra_server.erl
```diff
@@ -1290,6 +1304,13 @@ follower_catchup_cond(_, _Msg, State) ->
 wal_down_condition(_Msg, #{log := Log} = State) ->
     {ra_log:can_write(Log), State}.
 
+transfer_leadership_condition(#append_entries_rpc{term = Term},
```
kjnilsson: It is theoretically possible that the next message could be an `install_snapshot_rpc` so we'd need to handle this as well.

src/ra_server_proc.erl
```diff
@@ -634,6 +647,8 @@ follower(_, tick_timeout, State) ->
     {keep_state, State, set_tick_timer(State, [])};
 follower({call, From}, {log_fold, Fun, Term}, State) ->
     fold_log(From, Fun, Term, State);
+follower(_, transfer_leadership, State) ->
+    {next_state, candidate, State, [election_timeout_action(short, State)]};
```
kjnilsson: It would be quicker to issue the `vote_rpc`s here just like happens when a follower move to pre_vote or a pre_vote moves to candidate. Then the latency of this operation should be shorter. To do so it may be better to move the handling of this message to `ra_server` as that is where that code is currently executed.

PR #136, opened 2019-10-30
Handle duplicated entries in WAL in ra_dbg

acogoluegnes:


src/ra_dbg.erl
```diff
@@ -28,7 +31,7 @@ replay_log(WalFile, Module, InitialState) ->
 -spec replay_log(string(), module(), term(), Fun) -> term() when
   Fun :: fun((State :: term(), Effects :: term()) -> term()).
 replay_log(WalFile, Module, InitialState, Func) ->
-  Wal = lists:reverse(ra_log_wal:wal2list(WalFile)),
+  Wal = filter_entries_duplicates(ra_log_wal:wal2list(WalFile)),
```
michaelklishin: How about `filter_duplicate_entries`?

PR #147, opened 2019-11-18
Allow aux and machine monitors to monitor same pid

kjnilsson:
At the same time. Monitoring refactoring.


kjnilsson:
Some failures in the rabbitmq test suites. Investigating.

kjnilsson:
Fixed. Ready for review.

Issue #153, opened 2019-12-17
Pre-allocate WAL file when using fdatasync


kjnilsson:
Currently it is very likely that using `file:datasync/1` in the wal always gets upgraded to `fsync` as the metadata will need to be updated because the file size has changed. If we pre-allocate WAL files we may be able to reap more benefits from using `fdatasync`.

To pre allocate we can try using:

```
file:allocate(Fd, 0, MaxSize),
file:position(Fd, MaxSize),
file:truncate(Fd),
```

kjnilsson:
The file mode will need to change from `append` to `write` as well

kjnilsson:
closing as merged.

PR #156, opened 2019-12-19
Pre allocate wal file

kjnilsson:
This should benefit use of fdatasync and direct io scenarios as
fdatasync may well be upgraded to full fsync when the file size is
extended (as it is for every write).

Referece: #153

### DIFF
--- a/src/ra_log_wal.erl
+++ b/src/ra_log_wal.erl
@@ -44,8 +44,6 @@
 
 -record(batch, {writes = 0 :: non_neg_integer(),
                 waiting = #{} :: #{pid() => #batch_writer{}},
-                                   % {From :: ra_index(), To :: ra_index(),
-                                   %  Term :: ra_term()}},
                 start_time :: maybe(integer()),
                 pending = [] :: iolist()
                }).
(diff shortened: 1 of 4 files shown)


src/ra_log_wal.erl
```diff
@@ -527,12 +529,29 @@ open_wal(File, Max, #conf{write_strategy = o_sync,
...
     {ok, Fd} = ra_file_handle:open(File, Modes),
+    maybe_pre_allocate(Fd, Max),
     ok = write_header(Fd),
     ok = ra_file_handle:sync(Fd),
     {Conf, #wal{fd = Fd,
                 max_size = Max,
                 filename = File}}.
 
+maybe_pre_allocate(Fd, Max) ->
+    case ra_env:wal_pre_allocate() of
```
kjnilsson: This does a call into the configuration process, can we get this value at init time and stash it in the `conf` record?

src/ra_log_wal.erl
```diff
@@ -527,12 +529,29 @@ open_wal(File, Max, #conf{write_strategy = o_sync,
...
+        true ->
+          %% extend wal file
+          ok = pre_allocate(Fd, Max);
+        _ ->
+          no_pre_allocate
```
kjnilsson: If we're not matching on the return then just return `ok` here

PR #160, opened 2020-01-07
Move commit latency to microseconds

kjnilsson:
Default to 0
Also reset after recovery to ensure really large incorrect values are
calculated.

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -308,7 +308,9 @@ recover(#{id := {_, _, LogId},
                  end,
                  State0, []),
     Log = ra_log:release_resources(1, Log0),
-    State#{log => Log}.
+    State#{log => Log,
+           %% reset commit latency as recovery may calculate a very old value
+           commit_latency => 0}.
 
 -spec handle_leader(ra_msg(), ra_server_state()) ->
     {ra_state(), ra_server_state(), effects()}.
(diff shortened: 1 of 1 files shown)


gerhard:
This is a reminder to do the following when we bump to a `ra` version that has this change:

- [ ] [Rename `raft_entry_commit_latency` to `raft_entry_commit_latency_seconds`](https://github.com/rabbitmq/rabbitmq-prometheus/blob/dd9f8f562b2017dc9ff7aaeb52cba055d89a0e4b/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl#L159)
- [ ] [Convert from micros to seconds](https://github.com/rabbitmq/rabbitmq-prometheus/blob/dd9f8f562b2017dc9ff7aaeb52cba055d89a0e4b/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl#L193)

kjnilsson:
The idea was that we'd lock to hex releases for actually releasable branches and keep the master branch using Ra master for early integration.

gerhard:
Are we planning on cutting a new ra release & shipping in the upcoming RabbitMQ v3.8.3?

I'm asking because I want to back-port this to https://github.com/rabbitmq/rabbitmq-prometheus/commit/a363a92a59c4c43bf6d7f74e6ee32d71455d8215 `v3.8.x` but cannot do it without a bump to ra: https://github.com/rabbitmq/rabbitmq-server/blob/v3.8.3-beta.3/rabbitmq-components.mk#L119

kjnilsson:
I just realised that this was reverted due to being a breaking change.

gerhard:
OK, so we are keeping milliseconds for the unit, but resetting it correctly.

Thumbs up if we expect to ship ra v1.0.6 in RabbitMQ v3.8.3.

PR #174, opened 2020-03-31
Recover wal chunks

Vanlightly:
## Proposed Changes

Rather than read each WAL in its entirety, read it in chunks, in order to reduce the memory footprint during recovery.

No breaking changes.

### DIFF
--- a/src/ra_log_wal.erl
+++ b/src/ra_log_wal.erl
@@ -268,9 +268,7 @@ recover_wal(Dir, #conf{segment_writer = TblWriter,
     All = [begin
 
                Fd = open_at_first_record(F),
-               ok = recover_wal_chunks(Fd, RecoveryChunkSize, #{
-                        chunk_remainder => <<>>, 
-                        chunk_remainder_len => 0}),
+               ok = recover_wal_chunks(Fd, RecoveryChunkSize, #{}),
                close_existing(Fd),
                recovering_to_closed(F)
            end || F <- WalFiles],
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@Vanlightly Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/ra&pullRequestId=174)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/ra&pullRequestId=174) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@Vanlightly Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/ra&pullRequestId=174)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/ra&pullRequestId=174) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@Vanlightly Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/ra&pullRequestId=174)!

src/ra_log_wal.erl
```diff
@@ -714,46 +740,80 @@ dump_records(<<_:1/unsigned, 1:1/unsigned, _:22/unsigned,
...
-recover_records(<<Trunc:1/unsigned, 0:1/unsigned, IdRef:22/unsigned,
+% Reached zeros indicating end of wal file
+recover_records(_, <<_:1/unsigned, 0:1/unsigned, _:22/unsigned, 
+                    IdDataLen:16/unsigned, _:IdDataLen/binary,
+                    _:32/integer, 0:32/unsigned, _/binary>>,
+                  _, _) ->
+    ok;
+% First record or different UID to last record
+recover_records(Fd, <<Trunc:1/unsigned, 0:1/unsigned, IdRef:22/unsigned,
                   IdDataLen:16/unsigned, UId:IdDataLen/binary,
```
kjnilsson: alignment

src/ra_log_wal.erl
```diff
@@ -714,46 +740,80 @@ dump_records(<<_:1/unsigned, 1:1/unsigned, _:22/unsigned,
...
-                    "Continuing.~n", [Err]),
-              ok
-    end.
+recover_wal_chunks(Fd, RecoveryChunkSize, Cache) ->
+    Chunk = read_from_wal_file(Fd, RecoveryChunkSize),
+    recover_records(Fd, Chunk, Cache, RecoveryChunkSize).
 
-recover_records(<<Trunc:1/unsigned, 0:1/unsigned, IdRef:22/unsigned,
+% Reached zeros indicating end of wal file
+recover_records(_, <<_:1/unsigned, 0:1/unsigned, _:22/unsigned,
```
kjnilsson: I think you can match 0 for all integers here, just to be explicit

src/ra_log_wal.erl
```diff
@@ -714,46 +740,80 @@ dump_records(<<_:1/unsigned, 1:1/unsigned, _:22/unsigned,
...
+                  Rest/binary>>,
+                Cache, RecoveryChunkSize) ->
     true = validate_and_update(UId, Checksum, Idx, Term, EntryData, Trunc),
-    % TODO: recover writers info, i.e. last index seen
-    recover_records(Rest,
-                    Cache#{IdRef =>
-                           {UId, <<1:1/unsigned, IdRef:22/unsigned>>}});
-recover_records(<<Trunc:1/unsigned, 1:1/unsigned, IdRef:22/unsigned,
+    recover_records(Fd, Rest,
+                    Cache#{
```
kjnilsson: This kind of map formatting isn't done anywhere else in this module so looks a bit out of place here.

src/ra_log_wal.erl
```diff
@@ -714,46 +740,80 @@ dump_records(<<_:1/unsigned, 1:1/unsigned, _:22/unsigned,
...
+                        IdRef => {UId, <<1:1/unsigned, IdRef:22/unsigned>>}
+                    }, RecoveryChunkSize);
+
+ % TODO: recover writers info, i.e. last index seen
+% Same UID as last record
+recover_records(Fd, <<Trunc:1/unsigned, 1:1/unsigned, IdRef:22/unsigned,
                   Checksum:32/integer,
```
kjnilsson: alignment

src/ra_log_wal.erl
```diff
@@ -714,46 +740,80 @@ dump_records(<<_:1/unsigned, 1:1/unsigned, _:22/unsigned,
...
+    end;
+% Reached the of chunk, with some remainder, time to read the next chunk
+recover_records(Fd, Chunk, Cache, RecoveryChunkSize) ->
+    NextChunk = read_from_wal_file(Fd, RecoveryChunkSize),
+    
+    case NextChunk of 
+        <<>> ->
+            % we were expecting more to be read and there wasn't meaning
+            % we have a partial record. This should never happen.
+            ?WARN("Partial record lost", []),
```
kjnilsson: This isn't necessarily the case as you'd arrive here if you had a partial zero-padded remainder which isn't a lost record

src/ra_log_wal.erl
```diff
@@ -714,46 +740,80 @@ dump_records(<<_:1/unsigned, 1:1/unsigned, _:22/unsigned,
...
+            <<>>;
+        {error, Reason} ->
+            exit({could_not_read_wal_chunk, Reason})
+    end.
 
 validate_and_update(UId, Checksum, Idx, Term, EntryData, Trunc) ->
     validate_checksum(Checksum, Idx, Term, EntryData),
     true = update_mem_table(ra_log_recover_mem_tables, UId, Idx, Term,
-                            binary_to_term(EntryData), Trunc =:= 1).
+                          binary_to_term(EntryData), Trunc =:= 1).
```
kjnilsson: alignment

src/ra_log_wal.erl
kjnilsson: Maybe lower to a debug level and prepend the message with "wal: " or something to make it clear where it is coming from.

src/ra_log_wal.erl
kjnilsson: Might be nicer to assign a new Cache variable on the line before

src/ra_log_wal.erl
Vanlightly: As far as I see, this should only match the situation where we have a partial record, and we're at the end of the file. A remainder cannot be zero padded (in the sense that the zeros indicate that we've reached the end of the data). A remainder can only exist if the chunk did not have the whole record, if it has zeros at the end then we have a bad data situation.

kjnilsson: Surely if you write records into a zero padded wal file and you happen to have a couple of empty bytes left at the end you'd reach this clause without it being an error?

kjnilsson: Ok, i've made some tweaks, I don't think it should be a warn as it may just worry users when it isn't a big deal, any reminder at the end of the file that cannot be parsed can safely be discarded

PR #176, opened 2020-04-03
Fix integer overflow bug in segment index

kjnilsson:
The file offset was for some inexplicable reason recorded as a 32 bit
integer effectively limiting max file size to ~4GB. With large Ra
commands and large segment max entries configured it is fairly easy to
go beyond this causing data in the segment to be overwritten.

This change stores the offset in 64 bit but remains backwards compatible
by incrementing the segment file version and handling this in the code.

### DIFF
--- a/src/ra_log_segment.erl
+++ b/src/ra_log_segment.erl
@@ -19,12 +19,13 @@
 
 -include("ra.hrl").
 
--define(VERSION, 1).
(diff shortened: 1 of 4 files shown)


kjnilsson:
See https://github.com/rabbitmq/ra/issues/175 for context

PR #183, opened 2020-05-11
ra: Fix force_delete_server/1 spec

dumbbell:
It can return `{badrpc, Reason}` from the underlying `rpc:call/4`.

Found by Dialyzer in rabbitmq-server.


dumbbell:
For the record, the error was reported by Dialyzer in rabbitmq-server:
```
rabbit_quorum_queue.erl:953: The variable Err can never match since previous clauses completely covered the type
          'ok' | {'error', _}
```

PR #184, opened 2020-05-15
clarifying spec fixes for queries

mbj4668:
## Proposed Changes

More spec fixes.  Previoulsy, most of the query functions were speced like:

                ra_server_proc:ra_leader_call_ret(term()) |
                {ok, {ra_idxterm(), term()}, not_known}

This is type-correct, but a bit confusing.  The term() in the first row is really {ra_idxterm(), Reply}.

## Types of Changes

- [x ] Bug fix (non-breaking change which fixes issue #NNNN)

## Checklist

- [x ] I have read the `CONTRIBUTING.md` document
- [x ] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ x] All tests pass locally with my changes

### DIFF
--- a/src/ra.erl
+++ b/src/ra.erl
@@ -768,7 +768,7 @@ pipeline_command(ServerId, Command) ->
 %% @end
 -spec local_query(ServerId :: ra_server_id(),
                   QueryFun :: query_fun()) ->
-    ra_server_proc:ra_leader_call_ret({ra_idxterm(), term()}).
+    ra_server_proc:ra_leader_call_ret({ra_idxterm(), Reply :: term()}).
 local_query(ServerId, QueryFun) ->
     local_query(ServerId, QueryFun, ?DEFAULT_TIMEOUT).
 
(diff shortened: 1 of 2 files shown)


mbj4668:
I'm sorry, I ran the wrong dialyzer tests...  I will fix this.

mbj4668:
Perhaps the documentation for local_query() and leader_query() can explain what the ra_idxterm() is and how it is useful.  Why isn't it returned from consistent_query()?

michaelklishin:
Reference docs generally should not explain concepts, in particular if it's of little use most of the time.

michaelklishin:
While I don't know if there is a specific reason, `consistent_query/{2,3}` is a version of `leader_query/{2,3}` in that it checks that the current leader still, in fact, is the leader. If that's the case you probably don't need the term information.

With a query type that can return stale data such as `local_query/2` you potentially need it to be able to determine if the data is stale. Since you are explicitly telling Ra to not go through the leader with extra consensus checks, it's up to your application to do that.

michaelklishin:
Ra needs a set of guides which I hope to be able to contribute to. 
Thank you!

PR #186, opened 2020-05-22
Allow the binary serialization function to be configurable

kjnilsson:
Primarily this will allow the use of term_to_iovec instead of
term_to_binary.


michaelklishin:
Does such MFA invocation with `iovec` still perform better than `term_to_binary/1`?

kjnilsson:
> Does such MFA invocation with `iovec` still perform better than `term_to_binary/1`?

That is a good question, it doesn't make it worse what I can see but @Vanlightly will take a look and run some comparative tests with either setting.

BTW we can't merge this yet as I didn't make the test suite OTP 21 and 22 compat!

PR #190, opened 2020-06-13
handle corrupt WAL files

tomyouyou:
Broker will start failed with corrupt WAL files, and write a log like:

09:41:02.487 [error] Supervisor ra_sup had child ra_system_sup started with ra_system_sup:start_link() at undefined exit with reason {shutdown,{failed_to_start_child,ra_log_sup,{shutdown,{failed_to_start_child,ra_log_wal_sup,{shutdown,{failed_to_start_child,ra_log_wal,{unknown_wal_file_format,<<"abec">>,10}}}}}}} in context start_error
09:41:02.487 [error] CRASH REPORT Process <0.229.0> with 0 neighbours exited with reason: {{shutdown,{failed_to_start_child,ra_system_sup,{shutdown,{failed_to_start_child,ra_log_sup,{shutdown,{failed_to_start_child,ra_log_wal_sup,{shutdown,{failed_to_start_child,ra_log_wal,{unknown_wal_file_format,<<"abec">>,10}}}}}}}}},{ra_app,start,[normal,[]]}} in application_master:init/4 line 138
{"Kernel pid terminated",application_controller,"{application_start_failure,ra,{{shutdown,{failed_to_start_child,ra_system_sup,{shutdown,{failed_to_start_child,ra_log_sup,{shutdown,{failed_to_start_child,ra_log_wal_sup,{shutdown,{failed_to_start_child,ra_log_wal,{unknown_wal_file_format,<<\"abec\">>,10}}}}}}}}},{ra_app,start,[normal,[]]}}}"}
Kernel pid terminated (application_controller)

### DIFF
--- a/src/ra_log_wal.erl
+++ b/src/ra_log_wal.erl
@@ -260,13 +260,16 @@ recover_wal(Dir, #conf{segment_writer = TblWriter,
     All = [begin
                FBase = filename:basename(F),
                ?DEBUG("wal: recovering ~s", [FBase]),
-               Fd = open_at_first_record(F),
-               {Time, ok} = timer:tc(fun () ->
-                                             recover_wal_chunks(Fd, RecoveryChunkSize)
-                                     end),
-               ?DEBUG("wal: recovering ~s took ~bms", [FBase, Time div 1000]),
-               close_existing(Fd),
-               recovering_to_closed(F)
+               case open_at_first_record(F) of
+                   error -> ok;
+                   Fd ->
+                    {Time, ok} = timer:tc(fun () ->
+                                                    recover_wal_chunks(Fd, RecoveryChunkSize)
+                                            end),
+                    ?DEBUG("wal: recovering ~s took ~bms", [FBase, Time div 1000]),
+                    close_existing(Fd),
+                    recovering_to_closed(F)
+                end
            end || F <- WalFiles],
     % get all the recovered tables and insert them into closed
     Closed = lists:append([C || {C, _, _} <- All]),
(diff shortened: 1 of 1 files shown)


michaelklishin:
@tomyouyou would you like to add a test case or two for such WAL segment files?

michaelklishin:
@kjnilsson mentioned that we do not want to delete WAL segment files automatically. In part this would make it nearly impossible to find out under what circumstances a certain file was produced.

@tomyouyou thank you for your ongoing contributions, though!

tomyouyou:
> @tomyouyou would you like to add a test case or two for such WAL segment files?

This issue has occurred in our  lab environment. The reason is that the ‘/ var /’  directory has been used up, causing WAL files to be damaged.

michaelklishin:
@tomyouyou in cases such as running out of disk space, such solution might be reasonable. However, automatically deleting segment files means we will no longer have *any* information to work with when troubleshooting other segment corruption cases. Maybe when Ra is a very mature, stable and barely changing library we would consider this but at the moment the upside is not worth the downside of this behavior.

Running out of disk space is an operational issue 101. It means that the system did not have sufficient monitoring in place. A lot of software won't work as expected and won't necessarily be able to recover from such condition. We always recommend decomissioning such nodes as this is not a given that every component that writes to disk would be able to recover and reconcile with its cluster peers. One day when the schema database is Raft-based as well it would be more feasible but we are not there yet.

@kjnilsson

tomyouyou:
@michaelklishin 
Our system has disk monitoring measures. If the disk has been used up, there will be corresponding self-healing measures, such as deleting some unimportant logs or backup files. When disk space becomes available again, all software should be able to self-healing without human participation.

For the software itself, the first consideration is the integrity of function, which is understandable. while the engineering application is more concerned with self-healing in any case. It is recommended to consider the following aspects of engineering application, such as adding a configuration option(Off by default), which can be self-healing after it is turned on.

michaelklishin:
> When disk space becomes available again, all software should be able to self-healing without human participation

the reality of distributed stateful systems is a little bit less rosy, however. Deleting data may seem "self-healing" to you but it would be catastrophic to others and would prevent our team from investigating other possible segment file corruption scenarios, which all software eventually runs into.

> while the engineering application is more concerned with self-healing in any case

you are making some bold statements here and assume that every user and operational team has exact same set of priorities. This is not the case, and the suggestion that we should bake this into Ra, a library, sounds a bit arrogant to me.

Deleting data only makes some sense when the disk is full, it is not necessarily the case. In other cases of unrecoverable segment files you **want to preserve the files** for inspection.

If you want to delete data to recover from a full disk, you can do it manually without making Ra do it. The whole point of Raft is to guarantee state machine transition log safety by replicating it. If one node fails, gets its local log wiped out and restarts, as long as the majority of nodes are up and clients are data safety-aware (e.g. they retry operations), you can decomission the node or wipe its log and restart it. It will catch up with the current leader.

I don't know how @kjnilsson feels about a way to opt-in for deletion. Both deletion and skipping such files leads us to the log gap detection class of problems, something that Raft does not really focus on with its "commit index-oriented" approach to things.

Issue #202, opened 2020-11-23
Avoid incrementing next_index when a append entries rpc is dropped


kjnilsson:
All replication messages sent by Ra are sent using `noconnect` and `nosuspend` which means messages may be dropped, especially when the distribution buffer fills up. This means that replication needs to go through the "skip back" procedure where a follower returns a failure and the server has to skip back and resend which can cause unnecessary degradation of performance.

Instead we could only increment next_index _after_ a successful rpc send. This does mean a few internal code changes but could well be worth the effort especially for scenarios where large amounts of data is transferred.

kjnilsson:
Done with a different approach.

PR #203, opened 2020-11-25
Retry send in separate process after nosuspend

kjnilsson:
Instead of dropping all rpcs and relying on the retry protocol we start
a separate process to retry the rpc send allowing it to block until the
send has completed after which it updates the peer's status so it can
start pipelining again.


kjnilsson:
Addresses https://github.com/rabbitmq/ra/issues/202

PR #222, opened 2021-06-09
Optimise ra_log_segment:init

kjnilsson:
In particular the call the filename:absname/1 which isn't actually needed
was very slow. This should result in a ~40% speedup for the read_opt test case.

This also removes the `read_ahead` option which would never have been used given the strategy to read entries backwards (to be reviewed).

The `read_opt` test in ra_log_2_SUITE shows a substantial improvement in log initialisation time taken.

AFTER:
read took 0.389ms Reduction 1704
read init took 10.937ms Reduction 151330

BEFORE:

read took 0.42ms Reduction 3063
read init took 17.449ms Reduction 238108

### DIFF
--- a/src/ra_log.erl
+++ b/src/ra_log.erl
@@ -48,11 +48,6 @@
 
 -include("ra.hrl").
 
--define(METRICS_CACHE_POS, 2).
--define(METRICS_OPEN_MEM_TBL_POS, 3).
--define(METRICS_CLOSED_MEM_TBL_POS, 4).
--define(METRICS_SEGMENT_POS, 5).
-
 -define(DEFAULT_RESEND_WINDOW_SEC, 20).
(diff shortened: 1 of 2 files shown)


src/ra_log.erl
```diff
@@ -1000,7 +995,7 @@ server_data_dir(Dir, UId) ->
 %%%% TESTS
 
 -ifdef(TEST).
--include_lib("eunit/include/eunit.hrl").
+% -include_lib("eunit/include/eunit.hrl").
```
michaelklishin, line 998: Can this be removed entirely?

michaelklishin (approved):
It would be nice to remove the include entirely instead of commenting it out.

kjnilsson:
> It would be nice to remove the include entirely instead of commenting it out.

yeah I'll remove it in the next PR

PR #226, opened 2021-06-17
Add local server state query support

dumbbell:
The new `locally_known_members/{1,2}` functions do the same thing as `members/{1,2}` except that the specified node handles the query instead of the leader.

In other words, if the queried node is not the leader, it will still handle the query, returning a possibly out-of-date information. This differs from `members/{1,2}` where the same request would be redirected to the leader, returning the up-to-date consistent information.


dumbbell:
I just addressed @kjnilsson (out-of-band) comments: the existing `members/{1,2}` functions are expanded to take the server ID as `{local, ServerId}` if the user wants a non-blocking call, instead of adding another dedicated function with a weird name.

PR #232, opened 2021-10-05
Update to latest otp minors in github actions

pjk25:
23.2 -> 23.3
24.0 -> 24.1


pjk25:
It appears the github action we use to setup erlang does not have these minors yet. I will leave this PR here and try re-running it in a week.

PR #241, opened 2021-11-10
Optimise file operations to avoid use of the file server

kjnilsson:
Avoid any file based functions that use the file server in favour of "raw" versions.

### DIFF
--- a/src/ra.app.src
+++ b/src/ra.app.src
@@ -1,6 +1,6 @@
 {application,ra,
              [{description,"Raft library"},
-              {vsn,"2.0.1"},
+              {vsn,"2.0.2"},
               {licenses,["Apache","MPL"]},
               {links,[{"github","https://github.com/rabbitmq/ra"}]},
               {modules,[]},
--- a/src/ra_lib.erl
+++ b/src/ra_lib.erl
@@ -39,8 +39,13 @@
          retry/2,
          retry/3,
          write_file/2,
-         lists_chunk/2
+         lists_chunk/2,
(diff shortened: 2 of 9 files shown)


src/ra.app.src
```diff
@@ -1,6 +1,6 @@
 {application,ra,
              [{description,"Raft library"},
-              {vsn,"2.0.1"},
+              {vsn,"2.0.2"},
```
michaelklishin, line 3: Should this be `2.0.3` since `2.0.2` has already shipped?

kjnilsson: No for some reason this commit went into the wrong branch. Main has 2.0.1

src/ra_lib.erl
```diff
@@ -337,6 +347,72 @@ lists_take(_N, [], Acc) ->
...
+    case prim_file:read_file_info(Dir) of
+        {ok, #file_info{type=directory}} ->
+            true;
+        _ ->
+            false
+    end.
+
+is_file(File) ->
+    case prim_file:read_file_info(File) of
+        {ok, #file_info{type = directory}} ->
```
michaelklishin, line 361: Just a note: this behavior is consistent with that of `filelib`: https://www.erlang.org/doc/man/filelib.html#is_file-1

michaelklishin:
One interesting side effect of this: the faster the (parallel) QQ recovery process gets, the higher will be peak file handle use on the node.

michaelklishin (approved):

PR #246, opened 2021-12-20
Set off_heap for ra_servers

kjnilsson:
Ra tends to implement centralised utilities such as data stores and thus
will benefit from setting the off_heap message_queue_data strategy.

This will avoid messages being scanned by the Gc before they have been "received"
as well as enables further fan-in scalability.


michaelklishin:
@kjnilsson do you have any data that demonstrates how effective this may be?

michaelklishin:
We have discussed the effectiveness of this on Slack. While `off_heap` is not free, we expect this to be a net positive, primarily around GC runs.

michaelklishin:
After a longer discussion with @lhoguin we decided that it probably won't matter much for RabbitMQ and, in fact, most use cases.

PR #250, opened 2022-01-27
Ensure default system take wal_data_dir into account.

kjnilsson:
Right now it always defaults to the main Ra data dir.


michaelklishin:
Are the value identical or can the path change for existing installations?

kjnilsson:
That is a good question. If a system has wal_data_dir set already it will use a different directory. But this issue would also have been there in reverse for anyone with wal_data_dir set that upgraded from RabbitMQ 3.8 to 3.9.

michaelklishin (approved):

PR #255, opened 2022-03-01
Sparse read

kjnilsson:
This implements a more efficient way of reading a list of non consecutive indexes from the Ra log using a new `ra_log:sparse_read` function. This can be used by RabbitMQ quorum queues to read messages due to be delivered to consumers in a faster manner.

There is also an optimisation where it will automatically read ahead if some of the indexes in the requested list form a consecutive run (as is likely to happen for RabbitMQ quorum queues also).

### DIFF
--- a/src/ra_log.erl
+++ b/src/ra_log.erl
@@ -17,6 +17,7 @@
          append_sync/2,
          write_sync/2,
          take/3,
+         sparse_read/2,
          last_index_term/1,
          set_last_index/2,
          handle_event/2,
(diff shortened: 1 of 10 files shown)


src/ra_log.erl
```diff
@@ -993,6 +1046,18 @@ incr_counter(#cfg{counter = undefined}, _Ix, _N) ->
...
     filename:join(Dir, Me).
+
+maps_with_values(Keys, Map) ->
+    lists:foldr(
+      fun (K, Acc) ->
+              case Map of
+                  #{K := Value} ->
+                      [Value | Acc];
+                  _ ->
+                      Acc
```
dumbbell, line 1057: Is it possible that a requested index has no corresponding entry?

src/ra_log_segment.erl
```diff
@@ -258,10 +261,62 @@ read(State, Idx, Num) ->
...
+        {I, I} ->
+            %% no run, no cache;
+            undefined;
+        {FstI, LastI} ->
+            {_, FstOffset, FstLength, _} = map_get(FstI, SegIndex),
+            {_, LastOffset, LastLength, _} = map_get(LastI, SegIndex),
+            %% The cache needs to be at least as large as the next entry
+            %% but no larger than ?READ_AHEAD_B
+            MaxCacheLen = FstLength + LastOffset + LastLength,
+            CacheLen = min(MaxCacheLen, max(FstLength, ?READ_AHEAD_B)),
```
dumbbell, line 316: I don't understand this computation. If you are about to read `(Max)CacheLen` starting from `FstOffset`, doesn't `LastOffset` already include `FstOffset + FstLength` already?

I mean, if the maximum length to read should go from `FstOffset` to `LastOffset + LastLength`, then I would use:
```erlang
MaxCacheLen = LastOffset - FstOffset + LastLength
```
`LastOffset - FstOffset` corresponding to the area between the two offset and thus includes `FstOffset`.

The current computation of `MaxCacheLen` seems to be too large to me.

I do understand that you want `CacheLen` to be at least `FstLength` bytes to make sure that the entry is read entirely. But the logic doesn't correpond to the intention, at least to me, even though the result is probably correct in the end (in terms of the number of bytes read). Making sure that at least `FstLength` are read should be in the computation of `CacheLen`.

I'm not sure my argument is clear :-)

src/ra_log_segment.erl
```diff
@@ -475,36 +534,35 @@ read_header(Fd) ->
             Err
     end.
 
-pread(#state{cfg = #cfg{access_pattern = random,
-                        fd = Fd}} = State, Pos, Length) ->
+%% TODO: avoid updating state each time
```
dumbbell, line 537: Is this comment still relevant?

src/ra_log_segment.erl
```diff
@@ -514,3 +572,13 @@ index_record_size(2) ->
     ?INDEX_RECORD_SIZE_V2;
 index_record_size(1) ->
     ?INDEX_RECORD_SIZE_V1.
+
+%% returns the first and last indexes of the next consecuitive run
```
dumbbell, line 576: Typo in "consecuitive"

src/ra_log.erl
kjnilsson: `sparse_read` is only used by the `log` effect which in turn is only emitted by state machines which should know the range of available indexes as they control when snapshots can be taken. So no it shouldn't be possible and we can probably remove this clause.

kjnilsson: We could introduce some code to make the failure more explicit as was the case for the `log` effect previously. It exited with `failed_to_read_index_for_log_effect`

src/ra_log_segment.erl
```diff
@@ -258,10 +261,65 @@ read(State, Idx, Num) ->
...
+        {FirstIdx, LastIdx} ->
+            {_, FstPos, _FstLength, _} = map_get(FirstIdx, SegIndex),
+            {_, LastPos, LastLength, _} = map_get(LastIdx, SegIndex),
+            %% The cache needs to be at least as large as the next entry
+            %% but no larger than ?READ_AHEAD_B
+            MaxCacheLen = LastPos + LastLength - FstPos,
+            %% read at least the remainder of the block from
+            %% the first position
+            MinCacheLen = ?BLOCK_SIZE - (FstPos rem ?BLOCK_SIZE),
+            CacheLen = max(MinCacheLen, min(MaxCacheLen, ?READ_AHEAD_B)),
```
dumbbell, line 319: What if `FstLength` is greater than `?READ_AHEAD_B`, but lower than `MinCacheLen`? Like `FstPos` is close to the end of the block. I guess you still want to use `FstLength` as the minimum somewhere in the computation.

dumbbell (approved):

PR #260, opened 2022-03-16
Notify effect refactor

kjnilsson:


kjnilsson:
Tests that previously triggered the bug have run over night without any incidents

PR #262, opened 2022-03-22
Add a github actions workflow that produces rabbitmq OCIs

pjk25:
with Ra from PRs in this repo

Such images are useful in testing for performance regressions


mkuratczyk (approved):
LGTM

PR #268, opened 2022-04-04
Install higher versioned snapshot

kjnilsson:
Fixes around snapshot installations in combination with machine versions.

Firstly leaders will try to avoid sending snapshots to followers that do not yet have the code for the newer machine versions.

Also we better handle snapshots installs with a higher machine versions.


mkuratczyk (approved):
I can't reproduce the problem anymore. Thanks!

PR #269, opened 2022-04-05
Make first server trigger leader election

ansd:
Previously, a local Ra server triggered leader election when starting a
new cluster.

This commit makes the first server in the list of servers to trigger
leader election.

This commit is necessary for quorum `queue-leader-locator` (https://github.com/rabbitmq/rabbitmq-server/pull/4442) set to
`random` or `least-leaders` where leaders can be non-local to the node
where `ra:start_cluster/2,3,4` is called.

This commit will break other `ra` clients that depend on the local node to become the leader when starting a new Ra cluster. They will need to put the local ID first when calling `ra:start_cluster/2,3,4`. Alternatively, we could add an option to the public API (determining whether local node or first node should trigger election).

### DIFF
--- a/src/ra.erl
+++ b/src/ra.erl
@@ -328,6 +328,7 @@ start_cluster(System, ClusterName, Machine, ServerIds)
 %% @param ClusterName the name of the cluster.
 %% @param Machine The {@link ra_machine:machine/0} configuration.
 %% @param ServerIds The list of ra server ids.
+%% The first server in the list attempts to become the leader.
 %% @param Timeout The time to wait for each server to start
 %% @returns
 %% `{ok, Started, NotStarted}'  if a cluster could be successfully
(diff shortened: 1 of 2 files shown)


michaelklishin:
I don't think the current API makes it easy to depend on what node becomes the leader first, so maybe we should bump the minor and not worry about that until there's evidence that an option would be justified?

kjnilsson:
Would it not make sense when the desired leader isn't the local one to make an rpc call in rabbitmq to start the cluster from the target node? Then there would be no need for changes in Ra.

kjnilsson:
Also it would automatically "handle" the failure case where the chosen leader node isn't available.

ansd:
>Would it not make sense when the desired leader isn't the local one to make an rpc call in rabbitmq to start the cluster from the target node? Then there would be no need for changes in Ra.

That's a good idea as well :) I'll try it out.

> Also it would automatically "handle" the failure case where the chosen leader node isn't available.

In https://github.com/rabbitmq/rabbitmq-server/pull/4442 nodes that are `running` and not drained are preferred as leaders.

kjnilsson:
I think it is still preferrable to rpc in rabbit to handle that particular case.

For Ra it makes sense for new cluster to trigger an election in the local leader. At least you know an election did get triggered.

ansd:
I tested the RPC approach. Works well so far 👍 Thank you. Therefore closing this PR for now.

PR #272, opened 2022-04-28
Optimise reads in ra_log_meta

kjnilsson:
Instead of reading from dets we can use the ets cache.

### DIFF
--- a/src/ra_log_meta.erl
+++ b/src/ra_log_meta.erl
@@ -67,9 +67,8 @@ handle_batch(Commands, #?MODULE{ref = Ref,
                     #{Id := Data} ->
                         Inserts0#{Id => update_key(Key, Value, Data)};
                     _ ->
-                        case dets:lookup(Ref, Id) of
-                            [{Id, T, V, A}] ->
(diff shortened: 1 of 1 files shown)


mkuratczyk:
Just to capture what happened here: we noticed that with many QQs, it was easy to overload `ra_log_meta` with requests to store `last_applied`. After running a workload like `perf-test -x 1000 -y 1000 -r 10 -c 10 -qq -qp qq-%d -qpf 1 -qpt 1000 -z 30`, `ra_log_meta`'s mailbox would have hundreds of thousands of messages that it'd process for the next few minutes. This flamegraph was captured while RabbitMQ was processing a backlog of such messages (no other activity at the time):

before-ra-pr-272

It's clear that the `dets:lookup/2` is the culprit. The same information can be retrieved from `ets`, which is what happened in this PR. After the change, we could no longer reproduce the problem and even with twice as many queues, the mailbox wouldn't go above a few thousands messages, that can be processed quickly.

There could be other optimizations in the future (eg. in Osiris, a similar task is performed through atomics) but this is a quick win.

PR #278, opened 2022-05-17
Use term_to_iovec instead of term_to_binary in WAL

kjnilsson:
This should result in fewer allocations inside the WAL as
less binary data will typically be copied when serialising each entry term.

Include two perf page-fault captures: 

`main` branch:
out_mem_main
this PR:
out_mem_pr


kjnilsson:
Here are the cpu comparisons. There isn't that much in it.

Main:
out_cpu_main

This PR:
out_cpu_iovec

PR #287, opened 2022-05-31
Make snapshot compression options configurable

kjnilsson:
Use the `snapshot_compression_options` application configuration. Set to `[]` to disable compression all together.

Lower default from {compressed, 9} to `compressed`.

Use term_to_iovec instead of term_to_binary


kjnilsson:
Using RabbitMQ whilst consuming 1M messages (which results in snapshots being taken) we have the following _before_ perf capture:
out_with_snapshot_comp

And after:

out_snapshot_comp


It is notable that the time spent in `longest_match` is no longer done. Also on my devbox I got a higher consumption rate whilst capturing with perf (75k msg/sec vs 65k msg/sec). This could be anecdotal however.

PR #292, opened 2022-06-20
Make segment checksumming configurable

kjnilsson:
As with large messages this has considerable cost and with use of
erlang external format serialisation there is a degree of validation
anyway.


mkuratczyk (approved):
LGTM. The impact heavily depends on the message size and other factors but 10-15% throughput increase can be expected.

tiny messages:
pr292

4000B messages:
pr292-4000B

PR #296, opened 2022-07-04
add start_cluster/1

lazedo:
## Proposed Changes

trying to upgrade ra library from previous outdated version where System wasn't part of ra library.
adding start_cluster/1 which defers to start_cluster/2 with `default` will help upgrade

## Types of Changes

- [X] Bug fix (non-breaking change which fixes issue #NNNN)

## Checklist

- [X] I have read the `CONTRIBUTING.md` document
- [X] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] All tests pass locally with my changes

### DIFF
--- a/src/ra.erl
+++ b/src/ra.erl
@@ -33,6 +33,7 @@
          consistent_query/3,
          ping/2,
          % cluster operations
+         start_cluster/1,
          start_cluster/2,
          start_cluster/3,
          start_cluster/4,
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

src/ra.erl
```diff
@@ -374,6 +375,12 @@ start_cluster(System, ClusterName, Machine, ServerIds, Timeout)
 %% If a cluster could not be formed any servers that did manage to start are
 %% forcefully deleted.
 %% @end
```
the-mikedavis, line 377: Should this be documented as deprecated? Other functions that use the default `SystemId` are marked as deprecated like `start_server/1`/`restart_server/1`/`stop_server/1`

michaelklishin: Then it would not make much sense to accept this PR, would it?

the-mikedavis: I think the change is ok as a fix since it restores backwards compatibility. But if we plan on removing the functions that don't specify `SystemId` in the long run, maybe we shouldn't accept this, eh?

michaelklishin: Let's accept this and deprecate

lazedo:
Thank You @michaelklishin and @the-mikedavis

PR #307, opened 2022-09-27
Revert notify effect processing change

kjnilsson:
order done in 773aa85e8ba79c2aed2f4eb766c97ff525d3e2ce

This introduces additional write latency that is dependent on batch size and other effect processing times. Also it doesn't provide strong enough consistency guarantees as it tells you nothing about the effect processing on any nodes but the leader. It is better to assume eventual consistency when it comes to effect processing anyway and design systems accordingly.


kjnilsson:
The downside I see from this reversal is that it restores the original order only for "notifys", i.e. commands send using `ra:pipeline_command`. Replies to `ra:process_command` calls still return _after_ machine effect processing.

I.e. pipelining and waiting for the applied notification isn't quite the same as doing a `ra:process_command`. I am not sure if this matters or not but it is inconsistent.

Issue #167, opened 2020-02-14
Unsafe recovery option


kjnilsson:
Provide an option to force a ra server to start as a single node cluster in the case where a quorum cannot ever be re-established and we just want to recover what is left.

michaelklishin:
We've had some discussion notes about this and compared what Ra has today to other Raft-based systems, namely etcd and Consul.

## Links

* [Permanent loss of quorum requires new cluster](https://etcd.io/docs/v3.5/op-guide/runtime-reconf-design/#permanent-loss-of-quorum-requires-new-cluster)
* [etcd Disaster recovery](https://etcd.io/docs/v3.5/op-guide/recovery/)

## etc

* Permanent loss of cluster quorum requires a new cluster, by design
* Data safety is quoted as the primary reason for that
* Node recovery is performed using a snapshot file or an existing node data directory
* A node can be ([forced to boot](--force-new-cluster)): cluster size shrinks to one
* Keyspace data can be preserved for a node started with new configuration
* A node can be forced to join a cluster and forget about its previous one

## Consul

* Cluster membership uses a gossip protocol
* Recreated failed nodes (e.g. replacement pods) must retain node identity, which is derived from the IP address
* A node can be [forced to boot (bootstrap)](https://learn.hashicorp.com/tutorials/consul/recovery-outage): cluster size shrinks to one
* A node can be [forcefully removed from the cluster](https://www.consul.io/commands/force-leave)
* One of the [recommendations](https://learn.hashicorp.com/tutorials/consul/recovery-outage) suggests recovering a single node by forcing it to boot with only one known cluster member, and making a number of brand new nodes join it

## What Do We Want to Have in Ra

* Forced boot option
* Other ideas are out of scope for now

kjnilsson:
This was done in #306

PR #308, opened 2022-09-27
Recovery mode in a new force_restart_server call/3

erlmachinedev:
## Proposed Changes

The raft protocol itself is built around the notion of a leader which handles cluster change requests and elected by the majority which forms the cluster.

There is a use case when the majority nodes is gone (during network split or outage) and we still need data back. Especially in the scenario when we have inter datacenter communication and our requirement is to keep the minority group running after the incident.

This PR introduces the next calls:

`force_restart_server(System, ServerId, FilterNodes)`
`force_restart_server(System, ServerId, FilterNodes, AddConfig)`      

Which indicate in the client codebase that cluster is forcefully reduced to the setup which is under `FilterNodes` inclusion list. We tried to make minimal adjustment to not break the general approach behind library code and Ra protocol itself.

The state machine is left deterministic and the all improvement is built behind  `filter_nodes` setting which is passed through mutable config during the restart. The overhead is minimal since `validate_cluster/1` is to be called 3 times: during the recover, receiving snapshot or processing cluster change command.   

Also this PR includes `enq_drain_recovery` in addition to the default `enq_drain_basic` test SUITE which plays the split and recovery scenario into nemesis.

The respective documentation is added (plus small fixes) and one more sanity test. 

## Types of Changes

- [x] Documentation (correction or otherwise)

## Checklist

- [x] I have added necessary documentation (if appropriate)

## Further Comments

P.S. We see quite fresh and interesting [improvement](https://github.com/rabbitmq/ra/pull/306) from @kjnilsson but still need multi DC setup running which would allow us to handle quorum queues in multiple node setup as well. 

Also we took into account this [issue](https://github.com/rabbitmq/ra/issues/167).

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -363,10 +363,8 @@ recover(#{cfg := #cfg{log_id = LogId,
            [LogId,  EffMacVer, MacVer, LastApplied, CommitIndex, After - Before]),
     %% disable segment read cache by setting random access pattern
     Log = ra_log:release_resources(1, random, Log0),
-
     FilterNodes = maps:get(filter_nodes, State0),
     Cluster0 = maps:get(cluster, State0),
-
     Cluster1 = case FilterNodes of undefined ->
                        Cluster0;
                    _ ->
@@ -376,7 +374,6 @@ recover(#{cfg := #cfg{log_id = LogId,
                                            Res
                                    end, Cluster0)
               end,
-
     State#{cluster => Cluster1,
            log => Log,
            %% reset commit latency as recovery may calculate a very old value
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -84,7 +84,8 @@
       query_index := non_neg_integer(),
       queries_waiting_heartbeats := queue:queue({non_neg_integer(), consistent_query_ref()}),
       pending_consistent_queries := [consistent_query_ref()],
-      commit_latency => 'maybe'(non_neg_integer())
+      commit_latency => 'maybe'(non_neg_integer()),
+      nodes => 'maybe'([node()])
      }.
 
 -type ra_state() :: leader | follower | candidate
@@ -201,7 +202,9 @@
                             install_snap_rpc_timeout => non_neg_integer(), % ms
                             await_condition_timeout => non_neg_integer(),
                             max_pipeline_count => non_neg_integer(),
-                            ra_event_formatter => {module(), atom(), [term()]}}.
+                            ra_event_formatter => {module(), atom(), [term()]},
+                            %% distribution setup
+                            nodes := 'maybe'([node()])}.
 
 -type config() :: ra_server_config().
 
@@ -236,6 +239,7 @@ init(#{id := Id,
        machine := MachineConf} = Config) ->
     SystemConfig = maps:get(system_config, Config,
                             ra_system:default_config()),
+    Nodes = maps:get(nodes, Config, undefined),
     LogId = maps:get(friendly_name, Config,
                      lists:flatten(io_lib:format("~w", [Id]))),
     MaxPipelineCount = maps:get(max_pipeline_count, Config,
@@ -333,7 +337,8 @@ init(#{id := Id,
       aux_state => ra_machine:init_aux(MacMod, Name),
       query_index => 0,
       queries_waiting_heartbeats => queue:new(),
-      pending_consistent_queries => []}.
+      pending_consistent_queries => [],
+      nodes => Nodes}.
 
 recover(#{cfg := #cfg{log_id = LogId,
                       machine_version = MacVer,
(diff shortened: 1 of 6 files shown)


kjnilsson:
Thanks for this. I've taken a look and my main concern is that this allows you to change the effective membership on a temporary basis only (unless you do a membership change at which point it ends up persisting the filtered cluster).

Say in a cluster A, B, C, D, E where D and E for a partitioned minority. You force boot them with a filtered cluster of [D, E] where [A, B, C] keep making independent progress. Run for a while and then connectivity is restored and then what happens? You now have two very different histories that cannot be merged (you cannot simply restart D, E without the filter) so you are going to have to manually pick one and delete any nodes with a different history and remove and re-add them.

Now the same applies to my Pr #306 but in this case the membership change is made permanent in a manually changed member that you then use to seed the rest of the cluster from. This could also be used to restore availability on a minority side.

The point is I don't see any benefit with having the membership temporary as there is no way to remove the filter safely once it is in place.

src/ra_server.erl
```diff
@@ -84,7 +84,8 @@
       query_index := non_neg_integer(),
       queries_waiting_heartbeats := queue:queue({non_neg_integer(), consistent_query_ref()}),
       pending_consistent_queries := [consistent_query_ref()],
-      commit_latency => 'maybe'(non_neg_integer())
+      commit_latency => 'maybe'(non_neg_integer()),
+      filter_nodes => 'maybe'([node()])
```
michaelklishin, line 88: It's not clear from the name what nodes are being filtered, or whether these nodes act as filters, or what kind of filtering will be performed.

src/ra_server.erl
```diff
@@ -2107,6 +2114,20 @@ make_cluster(Self, Nodes) ->
...
 
+validate_cluster(State) ->
+    Filter = maps:get(filter_nodes, State),
+    Cluster0 = maps:get(cluster, State),
+    case Filter of
+        undefined ->
+            Cluster0;
+        _ ->
+            maps:filter(fun ({Name, Node}, _) -> Res = lists:member(Node, Filter),
+                                                 Res orelse ?INFO("~p is filtered on node: ~s", [Name, Node]),
```
michaelklishin, line 2125: I find this log message to be confusing. "[machine name] is filtered on node […]" does not really tell the user about the consequences.

michaelklishin (changes requested):
If the goal is to list some node that the cluster won't contact on boot, we should rename the setting to something like "nodes expected to be unavailable".

@kjnilsson and I had a different approach in mind but never finished it. I wouldn't say it was drastically different so we can see if this PR can be polished to a mergeable state.

kjnilsson:
> If the goal is to list some node that the cluster won't contact on boot, we should rename the setting to something like "nodes expected to be unavailable".
> 
> @kjnilsson and I had a different approach in mind but never finished it. I wouldn't say it was drastically different so we can see if this PR can be polished to a mergeable state.

I am reluctant to merge this without further discussing the issues I outlined in my comment above.

luos:
Thank you for reviewing these changes. 

We will definitely change any kind of naming to be a better one as needed. :)

Our primary goal is to recover service with as many messages recovered as possible, after it was made sure that the nodes lost can not be recovered. That is, we have a side channel to verify that those nodes are forever lost. After the cluster is recovered from minority, we would like to also grow the cluster back to the original size (may or may not be with the original nodenames). 

It would be good to be able to restore from any number of minority nodes. In a 5 node cluster this may mean 1 or 2 nodes. 
The reason we'd like to have 2 node recovery is that it is possible for the Followers to be behind the Leader in different amounts, so we'd like to give it the best chance to recover as much data as possible. 

The filter_nodes parameter in this PR expects the list of nodes which are considered still part of the cluster. The thinking here is that from RabbitMQ we can provide the list of nodes we still consider cluster members. In conjunction with `forget_cluster_node` therefore we can forget nodes which are lost. This way we can rely on which nodes mnesia considers cluster members. 

I checked #306 out as well, and I think I got what you are trying to do there. I think that could also be changed to accept a list of valid nodes, therefore electing the most up to date Follower, though as I understand, for that we would need to keep the commit index and similar data instead of resetting. The reason we did not really want to do this is that we have to coordinate the changes with all "up" nodes somehow, and in this PR we rely on mnesia for that. I am guessing we would need to have a multi step procedure of setting new cluster members on all peers, then force election.

What we were also going for is easy usage from RabbitMQ. As you can see, you can just forget_cluster_node each down node, then restart the nodes with this change, recovering the queues. This should (in theory) protect against "lost" nodes rejoining the cluster, though in our testing we ran into some issues with that - and nodes rejoined the mnesia cluster automatically. 

You are correct that this change would "remember" these lost nodes forever. We could not come up with a good solution to that. Maybe we can have a command to force a cluster member change into the ra log? We would like to ask your opinion and would be happy to implement. We'd prefer if the lost node would be forgotten.

In our testing if a node which was "lost" came back, definitely some weird behaviour of reelecting leaders were seen. I think the cluster will need to reject these peers in some way. 

A drawback of our changes is that it requires a restart - but we think that is acceptable in this scenario. 
Another may be that the filter_nodes param is pre-set on restart, so it should be only set after the cluster is fully built.

luos:
Hi @kjnilsson , do you have any suggestions on how to proceed on this?

Maybe we could lift your strategy of setting the cluster nodes with a call, but instead of resetting the current state, we could keep it? That would make it possible to elect the most up to date member.

kjnilsson:
Closing as we already have a recover mode that works ok.

PR #312, opened 2022-09-29
Make ra_systems_sup restart intensity configurable for tests

lhoguin:
Just a small change so Ra systems can be killed repeatedly in quorum_queue_prop_SUITE in companion PR https://github.com/rabbitmq/rabbitmq-server/pull/4167

### DIFF
--- a/src/ra_systems_sup.erl
+++ b/src/ra_systems_sup.erl
@@ -37,7 +37,11 @@ start_system(#{name := Name,
 
 
 init([]) ->
-    SupFlags = #{strategy => one_for_one, intensity => 1, period => 5},
+    %% This is not something we want to expose. It helps test suites
+    %% that crash Ra systems on purpose and may end up crashing
+    %% the systems faster than we normally allow.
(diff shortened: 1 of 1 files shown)


kjnilsson:
This is fine but I have to admit I don't know what a good default for this would be anyway. Based on your testing do you have any insight?

michaelklishin:
Given the importance of Ra in most systems, I'd say `{1, 5000}` sounds like a safe default. We have something similar in the Erlang client or federation to tolerate pretty mass client disconnects or failures.

The intent in tests is likely to lower it, not increase, am I right @lhoguin?

lhoguin:
Depends on the tests I suppose. The quorum_queue_prop_SUITE does a lot of crashing because it wants to check the queue recovers fine regardless of where it crashed, so it needs a high value. Other tests may want to allow zero crashes to make sure they're detected.

The current default is fine, it's also the supervisor's. `period` is in *seconds* so that is 1 restart allowed every 5 seconds.

michaelklishin:
@kjnilsson any objections to this default?

PR #313, opened 2022-10-03
Issue fadvice(DONT_NEED) after fsync

kjnilsson:
to tell kernel the data will not need to be read and the page cache can be reclaimed.


kjnilsson:
Closing as included properly in #316

PR #318, opened 2022-10-13
Use the rabbitmq bzlmod registry for erlang packages

pjk25:
Use https://github.com/rabbitmq/bazel-central-registry/tree/erlang-packages for erlang packages


pjk25:
Now that we have gazelle, I don't think the registry is worth the effort at this time

PR #323, opened 2022-11-04
Restore process_command/3 backwards compatibility

kjnilsson:
As older versions of Ra cannot understand the new reply mode format we use the old format whever a reply_mode is not specificed.

This also has smaller binary data overhead.


kjnilsson:
I'm going to do an early merge of this as it is needed to fix an regression in RabbitMQ. @the-mikedavis can always revert the PR if changes are needed or we just fix foward.

the-mikedavis (commented):
Ah whoops sorry for the breakage! This fix looks good to me. I won't need to revert it: sending `await_consensus` by default works 👍

PR #327, opened 2022-11-08
Minor performance improvement

ansd:
As Maxim Fedorov noted, there is no need to match against the table key when doing an ets:lookup/2 as the table key has to match anyways.

(I haven't done any measurements.)


kjnilsson (approved):

kjnilsson:
I doubt it makes any notable difference but let's have it!

PR #328, opened 2022-11-09
Write Ra server config async

kjnilsson:
To avoid blocking the ra server init call longer than necessary.


mkuratczyk:
Import times on my single-node Linux test box:
* 10k QQs, 60s before this change, 42s after
* 50k QQs, 302s before, 223s after

So a nice ~25% improvement, although it's still considerably slower than CQs (20s for 50k) or Streams (60s for 50k).

PR #329, opened 2022-11-09
Optimise RA server init

kjnilsson:
Dont write index 0 synchronously and don't fsync the config file to reduce blocking in gen_state:init function.


mkuratczyk:
I see ~50% improvement compared to `main`:

* 10k QQs, 60s before this change, 30s after
* 50k QQs, 302s before, 152s after

For reference, 50k CQs takes 20s, 50k Streams takes 60s on the same machine

PR #333, opened 2022-11-15
Perform Ra server initialisation outside of init callback.

kjnilsson:
To avoid creating a supervisor backlog when many Ra cluster are restarted and/or created in parallel.

The server start will still only return after the initialisation has been done however no work is done in the init callback unless the start happens from an old version of Ra.

### DIFF
--- a/src/ra_server_proc.erl
+++ b/src/ra_server_proc.erl
@@ -17,6 +17,7 @@
 
 %% State functions
 -export([
+         post_init/3,
          recover/3,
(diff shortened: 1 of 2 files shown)


mkuratczyk:
Importing 50k QQs, first using `main`, then PR333. From 313s to 109s to complete the import. Much higher CPU usage obviously. Memory usage remains the same. Great stuff!

import-50k-QQs

PR #341, opened 2022-12-15
Fixes for directories with unicode

kjnilsson:


michaelklishin:
Tested with directory names in Spanish and Russian, all tests pass (unlike in `main`).

michaelklishin (approved):

PR #342, opened 2022-12-16
Fix stale consistent read result after server restart

kjnilsson:
Reset cluster_change_permitted when becoming leader

As if there had been a leader change before this could reset to true
even before the leader has fully applied it's `noop` command for the new
term. This could result in occasional stale consistent read results.


kjnilsson:
I ran the `consistent_query_after_restart` test in a loop for 20 minutes without failures. Before the fix I could get it to fail every 5 runs or so.

michaelklishin (approved):

PR #350, opened 2023-02-03
Make segment_max_pending configurable in system.

kjnilsson:
This parameter is not configurable and it is questionable whether the default of 1024 is a good one.


kjnilsson:
Merging as uncontroversial

PR #353, opened 2023-02-17
Increase mixed machine version election availability.

### LINKED ISSUE #256: Allow lower versioned members to vote for higher versioned members if the effective machine version matches
Currently a member with a lower machine version will never vote for a member with a higher machine version. This is to avoid a higher versioned member being elected before there is a quorum of higher versioned members. This is somewhat too restrictive in the edge case where there already is a quorum of higher versioned members but the quorum is lost. 

Say if node A (leader) and B run version 2 and node C run version 1 and A is terminated currently neither B nor C can be elected leader. To allow C to vote for B we could allow this _if_ the effective machine version known by C matches the machine version of B.

To do so we can add a clause for this in the `ra_server:process_pre_vote` function to allow this specific case.

kjnilsson:
By allowing lower versioned members to vote for higher versioned members _if_ the effective machine version matches

This enables a cluster of, e.g. A(2), B(2), C(1) where A is the leader to still elect B as leader should A become unavilable.

Fixes #256


kjnilsson:
I've run some quorum queue mixed versions tests with this branch and it passes.

PR #361, opened 2023-03-15
Sunge/handle status callback

SimonUnge:
Very much a DRAFT. @kjnilsson

### DIFF
--- a/src/ra_machine.erl
+++ b/src/ra_machine.erl
@@ -71,7 +71,11 @@
 -export([init/2,
          apply/4,
          tick/3,
+<<<<<<< Updated upstream
          eval_members/7,
+=======
+         eval_members/6,
+>>>>>>> Stashed changes
          state_enter/3,
          overview/2,
          query/3,
@@ -201,7 +205,11 @@
               command_meta_data/0]).
 
 -optional_callbacks([tick/2,
+<<<<<<< Updated upstream
                      eval_members/6,
+=======
+                     eval_members/5,
+>>>>>>> Stashed changes
                      state_enter/2,
                      init_aux/1,
                      handle_aux/6,
@@ -236,10 +244,9 @@
                         ra_server_id(),
                         [ra_server_id()],
                         MacState :: state(),
-                        node(),
                         nodeup | nodedown |
-                        {add_member_result, Result} |
-                        {remove_member_result, Result}) -> effects().
+                        {add_member_result, Id, Result} |
+                        {remove_member_result, Id, Result}) -> effects().
 
 -callback init_aux(Name :: atom()) -> term().
 
@@ -293,12 +300,18 @@ tick(Mod, TimeMs, State) ->
                     ra_server_id(),
                     [ra_server_id()],
                     MacState :: state(),
-                    node(),
                     nodeup | nodedown |
+<<<<<<< Updated upstream
                     {add_member_result, Result} |
                     {remove_member_result, Result}) -> effects().
 eval_members(Mod, RaftState, Leader, Cluster, State, Node, Status) ->
     ?OPT_CALL(Mod:eval_members(RaftState, Leader, Cluster, State, Node, Status), undefined).
+=======
+                    {add_member_result, Id, Result} |
+                    {remove_member_result, Id, Result}) -> effects().
+eval_members(Mod, RaftState, Leader, Cluster, State, Status) ->
+    ?OPT_CALL(Mod:eval_members(RaftState, Leader, Cluster, State, Status), undefined).
+>>>>>>> Stashed changes
 
 %% @doc called when the ra_server_proc enters a new state
 -spec state_enter(module(), ra_server:ra_state() | eol, state()) ->
--- a/src/ra_machine.erl
+++ b/src/ra_machine.erl
@@ -71,6 +71,7 @@
 -export([init/2,
          apply/4,
          tick/3,
+         handle_status/8,
          state_enter/3,
          overview/2,
          query/3,
@@ -200,6 +201,7 @@
               command_meta_data/0]).
 
 -optional_callbacks([tick/2,
+                     handle_status/7,
                      state_enter/2,
                      init_aux/1,
                      handle_aux/6,
@@ -229,6 +231,8 @@
 -callback state_enter(ra_server:ra_state() | eol, state()) -> effects().
 
 -callback tick(TimeMs :: milliseconds(), state()) -> effects().
+
+-callback handle_status(any(), any(), any(), any(), any(), any(), any()) -> effects().
 
 -callback init_aux(Name :: atom()) -> term().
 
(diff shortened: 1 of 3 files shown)


SimonUnge:
Example implementation:
```
handle_status(leader, {ClusterName, _} = Leader, Cluster, _State, Node, nodeup) ->
    %% Figure out if we should add or do nothing                                                                                                                                             
    Conf = make_conf(ClusterName, {ClusterName, Node}, Cluster),
    [{add_member, Conf, {ClusterName, Node}, Cluster}];
handle_status(leader, {ClusterName, _} = Leader, Cluster, _State, Node, nodedown) ->
    %% Figure out if we should remove or do nothing                                                                                                                                          
    [{remove_member, {ClusterName, Node}, Cluster}];
handle_status(leader, _Leader, _Cluster, _State, ServerId, {What, Result}) when What == add_member_result;
                                                                                What == remove_member_result ->
    %% do something with the result...                                                                                                                                                       
    [];
```

src/ra_server_proc.erl
```diff
@@ -1188,6 +1230,81 @@ handle_effect(RaftState, {aux, Cmd}, EventType, State0, Actions0) ->
...
+    Me = self(),
+    %% Check if an member_eval pid is already ongoing. Do we need to monitor the pid?
+    Pid = case is_pid(OldPid) andalso is_process_alive(OldPid) of
+              true ->
+                  %% There is already a process working, do nothing
+                  OldPid;
+              false ->
+                  %% No process working, safe to spawn a new one.
+                  spawn(fun() ->
+                                Res = case ra:remove_member(Members, ServerId) of
```
kjnilsson, line 1282: I think these should only be handled by the `leader` state, if they ended up being triggered on a follower they should just be ignored. Also the first thing the temporary process should do is a consistent query into the leader just to check it is an "active" leader and not a partitioned one. `ra:remove_member/2` returns after the command has been added to the log, _not_ after consensus so it isn't quite enough.

src/ra_server_proc.erl
```diff
@@ -1188,6 +1230,81 @@ handle_effect(RaftState, {aux, Cmd}, EventType, State0, Actions0) ->
...
+                                                      {error, Err, R};
+                                                  Err ->
+                                                      {error, Err, R}
+                                              end;
+                                          {timeout, _} ->
+                                              {error, timeout};
+                                          E ->
+                                              E
+                                      end,
+                                ok = gen_statem:cast(Me, {remove_member, ServerId, Res})
```
kjnilsson, line 1301: I am thinking we can remove all the result handling and just reset the timer on monitor down.

src/ra_machine.erl
```diff
@@ -230,6 +232,14 @@
 
 -callback tick(TimeMs :: milliseconds(), state()) -> effects().
 
+-callback eval_members(ra_server:ra_state(),
```
kjnilsson, line 235: As mentioned below I don't think we need the result parts, also I don't think we need to know if a node was up down, this can all be derived from erlang APIs and isn't necessary to work out if we need to add a member or not. 

Can we reduce the `eval_member` API to just take the current list of members `[ra_server_id()]`, the leader server id server_id() and the MacState?

kjnilsson: Side note: adding `ra_state()` to the aux API was a mistake, the `ra_state()` is very much an internal concern which is why the aux state stuff isn't really documented. Can't change it now though. RA 3.0 maybe.

src/ra_server.erl
```diff
@@ -1879,6 +1890,14 @@ handle_down(RaftState, snapshot_writer, Pid, Info,
...
     {Log, Effects} = ra_log:handle_event({down, Pid, Info}, Log0),
     {RaftState, State#{log => Log}, Effects};
+handle_down(RaftState, member_eval, _Pid, _Info, State) ->
+    #{member_eval_pid := _OldPid} = State,
+    %% Q: Unsure if this handling is needed, we could in ra_server_proc
+    %% just check if the process is alive?
+    %% Q: No effect is returned here. We should perhaps return an effect that
+    %% sets the timer to a lower value. But, that is currently done when the
+    %% spawned process sends its result
+    {RaftState, State#{member_eval_pid => undefined}, []};
```
kjnilsson, line 1900: Set a short member_eval timeout here? Then we don't need any of the result handling.

src/ra_server_proc.erl
```diff
@@ -1188,6 +1230,81 @@ handle_effect(RaftState, {aux, Cmd}, EventType, State0, Actions0) ->
         handle_effects(RaftState, Effects, EventType,
                        State0#state{server_state = ServerState}),
     {State, Actions0 ++ Actions};
+handle_effect(_RaftState, {add_member, Conf, ServerId, Members}, _EventType,
```
kjnilsson, line 1233: `_RaftState` should be `leader` here to ensure only leaders evaluate this.

kjnilsson (changes requested):
Ok good progress. I think we can simplify further. Let's see how simple we can make this new API. :)

src/ra_machine.erl
kjnilsson: Correction: I don't even remember what the types are meant to be. :) `ra_state()` is the raft state!

Given this function should only be called by leaders I think it is still not required so can be removed.

src/ra_server_proc.erl
SimonUnge: Got it. How do I make sure its an active leader? Call to `leader_id` enough?

src/ra_server_proc.erl
SimonUnge: Done.

src/ra_server.erl
SimonUnge: Done

src/ra_server_proc.erl
SimonUnge: Done

src/ra_machine.erl
SimonUnge: Done

src/ra_machine.erl
```diff
@@ -230,6 +232,10 @@
 
 -callback tick(TimeMs :: milliseconds(), state()) -> effects().
 
+-callback eval_members(ra_server_id(),
+                       [ra_server_id()],
+                       MacState :: state()) -> effects().
```
SimonUnge, line 237: @kjnilsson Should we return list of effects, i.e `effects()`, or limit it to one effect, i.e `effect()`?

src/ra_server_proc.erl
```diff
@@ -1188,6 +1206,77 @@ handle_effect(RaftState, {aux, Cmd}, EventType, State0, Actions0) ->
         handle_effects(RaftState, Effects, EventType,
                        State0#state{server_state = ServerState}),
     {State, Actions0 ++ Actions};
+handle_effect(leader, {add_member, Conf, ServerId, Members}, _EventType,
```
SimonUnge, line 1209: Should we allow addition of multiple members at once? Should we allow adding and removing members as part of the same effect?

Issue #44, opened 2018-12-27
Non-voting members for stand-by backups


benoitc:
For a service I would like to have nodes that follow any events on the cluster without participating to the election. Main idea behind that is to create hot backups or simply nodes actings as readers. 

It seems that members of a cluster can already receive any events using the `state_enter/2` callback, so i'm thinking to implement it by adding a `watcher` role to a node when it is added to the cluster. The node would then receive any events from the replication but can't participate to the election. 

API would be similar to `add_member` with the addition of 3 functions: `add_watcher/{2,3}`, `remove_watcher/{2,3}`, `watchers/{1,2}`

Thoughts?

michaelklishin:
So in other words such nodes would permanently be followers?

I remember our team discussed something similar at some point. I don’t think RabbitMQ has any immediate uses for it but this can be useful.

benoitc:
> So in other words such nodes would permanently be followers?
exactly ^^

I will try to come with an implementation ASAP :)

benoitc:
@kjnilsson I was thinking about a non voter node that could potentially be promoted as a voter node. This had the following advantages:

* reuse the current replication logic, but the non voter follower would not declare itself as a candidate (by not triggering the election timeout)
- because it's using the whole replication, the latency will be the same
* eventually if needed a node could be promoted as a voter later. When for example some followers are down, or we cant to increase the number of replicated node or when we want to switch some machines. 

The change seems straight forward but i may miss some parts: It seems the  election timeout is handled  only in the `ra_server_proc` module an d may simply be done by modifying the state record and check for it in `maybe_set_election_timeout/2` and `election_timeout_action/2`

But **your proposal sounds more performant**. Maybe  promoting a node  to a full ra serve can also be done by simplifying the logic and done on demand ?  It would only require to have the whole log replicated. Like you're suggesting they may **not be full ra-servers but only processes that subscribe to the log and let one of the follower actually send it using the following process**:

	1. Subscription is sent to the leader
	2. The leader pick one of the follower and forward  the subscription to it
	4. the follower monitor the  watcher
	3. the follower acknowledge the subscription
	4. The follower ID is returned to the subscriber so it can monitor it

**Once ACKed the follower will start to send the logs and the watcher enter in a recovery phase**. If the subscription failed  (because the follower came down or maybe refused it), the watcher will get a return and may retry the subscription to not  block the leader.

Thoughts?

### Questions: 
- what happen when a new member is added? How does it get the latest state?

[...]

PR #367, opened 2023-04-28
WIP: Implement passive members for a ra cluster

luos:
## Proposed Changes

Passive members are not voting for any members.
Passive members are ignored for consensus purposes. Passive members are replicating all traffic from the leader. Passive members can be forced to be active members in the cluster.

Not handled yet:
Active members demoted to passive.
Handling Active/Active collisions a.k.a. split-brains. Some smaller changes marked as TODOs.

Working towards issue #44.

## Types of Changes

## Checklist

- [ ] I have read the `CONTRIBUTING.md` document

### DIFF
--- a/src/ra.erl
+++ b/src/ra.erl
@@ -59,12 +59,14 @@
          %% membership changes
          add_member/2,
          add_member/3,
+         add_member/4,
          remove_member/2,
          remove_member/3,
          leave_and_terminate/3,
          leave_and_terminate/4,
          leave_and_delete_server/3,
          leave_and_delete_server/4,
+         force_change_passive_members/3,
          %% troubleshooting
          % deprecated
          overview/0,
@@ -479,6 +481,7 @@ start_server(System, ClusterName, {_, _} = ServerId, Machine, ServerIds)
              id => ServerId,
              uid => UId,
              initial_members => ServerIds,
+             initial_passive_peers => [],
              log_init_args => #{uid => UId},
              machine => Machine},
     start_server(System, Conf).
@@ -564,7 +567,8 @@ delete_cluster(ServerIds, Timeout) ->
     {error, already_member} |
     {error, cluster_change_not_permitted}.
 add_member(ServerLoc, ServerId) ->
-    add_member(ServerLoc, ServerId, ?DEFAULT_TIMEOUT).
+    add_member(ServerLoc, ServerId, active, ?DEFAULT_TIMEOUT).
+
 
 %% @doc Same as `add_member/2' but also accepts a timeout.
 %% @see add_member/2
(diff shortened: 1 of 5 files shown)


illotum (commented):
You clearly spent more time on this, so most of my comments will be some form of questions:

It feels like the Roles and States overlap produces weird corner cases. Would it be possible to either extend `cluster` map to hold the passive role, or remove passive nodes from that map altogether? Or any other way to avoid storing pre_vote+passive or candidate+passive in the state?

I'd be happier if there was some thought given to auto-promotions. `passive_peers` being a list will be rough when it comes to adding another role, "passive but promote when conditions apply". Do you have plans, perhaps without an extra role, for auto-promotion?

Not clear about the intent behind `force_change_passive_memebrs`. State changes should go to the leader first, no?

luos:
Hi,

Yes, definitely I can see both ways have some advantages. We did not want to change the `cluster` map because that is more ephemeral state and this way only the consensus part needs changes and nothing else. TBH, we started with that but gave up for ease of implementation. :-) We also don't want (or at least as the first round of changes) passive peers to be less of a peer than active peers - except that they don't participate in elections.

We decided against introducing a new `statem` state to avoid having to touch the state machine very deeply and introducing any potential missed code paths. 

In this change passive peers are never promoted to be leaders or to active followers automatically. It is not possible to automate with any kind of safety.  If passive peers would be promoted automatically during a network partition it would jeopardize the safety and consistency of the raft cluster.

Our change aims at a disaster recovery scenario when recovering the majority is not possible. In case of a DR scenario, though, technically it can happen that the former leader tries to rejoin the cluster which would lead to a split brain scenario. 

In such a split brain, there is no guarantee that the `term` of the old leader is lower than the newly promoted leader's. For example in a 6 node cluster, with the 3 active + 3 passive setup, a network partition could trigger a disaster recovery scenario so the 3 passive nodes are promoted to leader, with `term+1`. Let's say if we can't shut down the three former active nodes, they could go through two leader elections while separated in their own data center. It would increase their terms by 2, so when they try to rejoin, they would be the preferred partition, which is not what the operator desired during the DR procedure. 

Any kind of auto-promotion would handle this scenario which is not possible in a safe way, similarly as RabbitMQ's classic queue mirroring, which suffers from the same problem. 

In an ideal world, whenever an active node is demoted to passive, the active node must be deleted, which may or may not be possible during network partitions. 

Keeping in mind a disaster recovery scenario, there might not be any active leaders present, hence the usage of the `force_change_passive_members` function. It must be handled by the local node, even if it's passive and follower, unblock the election procedure, and elect a new leader. 

If the cluster is up and healthy, we can introduce a non-forced version of the command, which goes through the leader/consensus. We can include that in a future commit.

illotum:
> It is not possible to automate with any kind of safety. If passive peers would be promoted automatically during a network partition it would jeopardize the safety and consistency of the raft cluster.

Perhaps I explain my use case poorly. I want auto-promotion to avoid joining slow members to the quorum. A node would go through the "catch up" period as a passive follower, and be promoted to full member when its log is close to the leader. If we do it via configuration change identical to adding new cluster member -- the log append, confirms, and the rest -- promotion should be safe. And of course if there is no majority there should be no promotion. 

Corollary, I think forcing commands through a minority would be better served by a separate PR (and be more generic). Would be nice to be able to force arbitrary commands.

Overall, I'm no @kjnilsson, but Ra being a fundamental library I'd value versatility and correctness over diff size.

---

Not to discount your use case BTW, disaster recovery is a very welcome feature for us too! LMK if I can somehow help.

Giving it some thought: is it really necessary to grow the cluster _before_ recovering from disaster? Why not force configuration change to a cluster of one, including election and term tick, and then add fresh members to it? Looking at prior art, that's what etcd does to my understanding.

kjnilsson:
> Perhaps I explain my use case poorly. I want auto-promotion to avoid joining slow members to the quorum. A node would go through the "catch up" period as a passive follower, and be promoted to full member when its log is close to the leader. 

I think this is a very reasonable feature and something that just didn't get done originally. I would encourage a separate pull request for this only. It should just be a temporary state flag of the peer that is set by default. The tricky bit would be to determine what "close to the leader" means. I think it would be sufficient to record the current index at the time of adding the new member and when the member reaches that index it can be included in consensus calculations. The assumption is this will help when adding members to systems with very long logs.

The index needs to be recorded in the cluster change command meta data such that we can ensure that members aren't able to flip back and forth between being included in consensus and not (e.g. if a leader election takes place before they have reached the pivot index).

kjnilsson:
> Giving it some thought: is it really necessary to grow the cluster _before_ recovering from disaster? Why not force configuration change to a cluster of one, including election and term tick, and then add fresh members to it? Looking at prior art, that's what etcd does to my understanding.

This is the approach taken in https://github.com/rabbitmq/ra/pull/306

Requiring recovery from a single member simplifies things substantially and reduces the risk of ending up in incomprehensible cluster states.

luos:
Hi @kjnilsson , 

After reading your [comment](https://github.com/rabbitmq/ra/pull/367#issuecomment-1532630664), do you think there is any chance something like this could be implemented and accepted with multiple passive followers, as in this change?

PR #369, opened 2023-05-09
Fix snapshot installation CRC failure

### LINKED ISSUE #368: ra_log_snapshot fails CRC check on OTP 26
This is due to the fact that the snapshot meta data map isn't replicated as binary data but instead re-serialised on the receiver which means it may not calculate the same checksum over that data (as term_to_binary map representation isn't deterministic between OTP versions). As the snapshot replication includes binary data from files we really ought to try to perform checksum validation if at all possible.

Options:

1. Disable checksum validation completely for `ra_log_snapshot` (this won't work when sending a snapshot to an old member that still performs the validation).
2. Do 1. but create a new snapshot implementation that does the right thing. This would require careful management of snapshot module configuration as it isn't possible to have Ra clusters with different snapshot implementations.
3. Replicate the entire snapshot file so that it includes the original binary representation of the meta data (and thus the CRC check isn't subject to OTP variations in the term_to_binary output).
    - to do this the sender would need to discover dynamically whether the receiver has the new code that can handle the entire file contents rather than just the machine data portion. We already perform an rpc to discover machine version compatibility so we can include a check for this also.
    -  At the receiving end the new implementation could pattern match on the first data chunk received to see if it includes the magic value and version (`<<"RASN", Version:32/integer>>`). If so the sender also has the new code and all is well. If not included the sender has the old logic and the receiver falls back to the old behaviour of serialising the meta data map but _without_ performing checksum validation. Instead we have to assume the data is fine and calculate and write a new checksum value into the snapshot file. 
    - This approach still has the shortcoming that a snapshot sent to a member running the old code may still fail due to term_to_binary output differences but this can't completely be avoided.

kjnilsson:
Due to the unwise use of term_to_binary rather than using the original binary data in the replication of snapshot states.

This change:

Introduces a new optional ra_snapshot callback: context/0

This is called by the sending Ra leader node to discover context and capabilities of the receiver. In this case it is used to indicate if the receiver is capabable of receiving the entire snapshot file.

Receiving the entire file is the updated approach that ensures the CRC check will be done on the same binary data it was generated from.

If the receiver does not have the context/0 callback or does not indicate support the old approach of sending the deserialised metat data map and any data following that is used.

When a snapshot is received from an old node (i.e. _not_ including the entire file) the receiver will not validate the checksum (as it may fail due to differences in map serialisation) and instead patch up it's local file with it's own calculcated checksum.

The scenario where a snapshot taken by a newer version of OTP and is then sent to a member using the old code cannot be handled and the old node will fail at snapshot checksum validation.

Fixes #368

### DIFF
--- a/src/ra.hrl
+++ b/src/ra.hrl
@@ -39,7 +39,10 @@
 %% after node restart). Pids are not stable in this sense.
 -type ra_server_id() :: {Name :: atom(), Node :: node()}.
 
--type ra_peer_status() :: normal | {sending_snapshot, pid()} | suspended.
+-type ra_peer_status() :: normal |
+                          {sending_snapshot, pid()} |
+                          suspended |
+                          disconnected.
 
 -type ra_peer_state() :: #{next_index := non_neg_integer(),
                            match_index := non_neg_integer(),
--- a/src/ra_log_snapshot.erl
+++ b/src/ra_log_snapshot.erl
@@ -15,11 +15,12 @@
          begin_accept/2,
          accept_chunk/2,
          complete_accept/2,
-         begin_read/1,
+         begin_read/2,
          read_chunk/3,
          recover/1,
          validate/1,
-         read_meta/1
+         read_meta/1,
+         context/0
          ]).
 
 -define(MAGIC, "RASN").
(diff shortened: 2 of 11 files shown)


michaelklishin (approved):

michaelklishin (approved):

michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/8143 makes rolling upgrades to Erlang 26 succeed under a constant load involving QQs.

PR #375, opened 2023-05-15
Non-voters and automatic promotion

illotum:
~~This is a draft.~~

My take on #44, mostly a continuation of design discussed there. Some remaining (low-stakes) questions are left _in italics_, but they will not change much.

Notably, a departure from previous designs, the feature is completely opt-in. Both server start and member addition need be extended according to this example:

```
UId = ra:new_uid(ClusterName),
New = #{id => Id, membership => promotable, uid => UId},
ok = ra:start_server(default, ClusterName, New, add_machine(), [ServerRef]),
{ok, _, _} = ra:add_member(ServerRef, New),
```

## Voter status

Voter status is stored in the cluster map of the server state and is part of every `$ra_cluster_update`. Additionally, nodes store their own status at the top level for ease of matching.

Nodes also store their own satus in `ra_state` ETS table **(breaking change)**, and present in overview.

## Effect on Raft

Voters behave as before. 

Non-voters ignore election timeouts and do not attempt to vote for others. Consensus calculation takes this into account, and non-voters have no effect on quorum size.

## Transitions

`ra:start_cluster` is unchanged, for new cluster of nonvoters will never elect a leader.

`ra:add_member` and `ra:start_server` now accept a map of server id and desired voter status. Legacy calls default to adding new server as a voter. 

On every `#append_entries_reply` leader may choose to promote non-voter by issuing new `$ra_join` with desired voter status. Currently, only one promotion condition is implemented: non-voter will be promoted when it reaches the leaders log index at the time of joining.


kjnilsson:
This looks very promising. Initial thought is that we shouldn't need a new command for this and that this should be the behaviour of all `ra_join` commands. New commands, although ok, can't be understood by old members if changes are made during a live upgrade so it is preferable to avoid adding new API unless absolutely necessary.

illotum:
Update after talking to @kjnilsson elsewhere. Removes voter promotion from the log, instead it is based on a single static target the nonvoter has to reach. Non-voters behave like regular followers, but are treated differently:

1. Leader doesn't count non voters when it computes required majority to raise commit index.
2. Pre vote requests from non-voters are ignored.
3. Pre vote and election responses from non voters are allowed (there's no way to tell who the ack is from). For this reason they also count towards required majority.

`ra:member_overview` reports a list of "voters" if requested from the leader.

luos:
Thank you @illotum for working on this. Seems we will be able to add our stuff on top of it when time comes. :) 

One thing not clear for me, is why is there a need for automatic promotion? For a RabbitMQ use case, I would imagine the benefits to be not that big as the cluster does not change very often?

illotum:
Taking the WIP off. It is unlikely polished, but deserves a review.

Some notes and concerns:

Voter state is part of the Raft log, within the cluster membership record. It is also tracked at a higher level purely for convenience in handling RPCs. This is a fairly superficial improvement compared to the effort and risk of confusion. Totally willing to forgo it.

Non-voting followers do not attempt to get elected, and ignore calls for election. All done within `ra_server` limits. Certain election timeouts are still kicked off in `ra_server_proc`, which is a waste.

Their match_index is not included in commit computation, yet they are reported in cluster membership. This may be confusing to consumers, e.g. when newly added server doesn't block progress at first.

Voter status is reported in `overview`. My biggest concern, as RabbitMQ instead directly reads `ra_state` table when it reports quorum critical. Issuing overview RPCs feels prohibitively expensive in comparison.

Promotion rules are more complicated than your design @kjnilsson, but not too far from it. This was just too low-hanging fruit too ignore. Non-voters that fail to get promoted remain passive followers until their state is handled manually.

Overall, most of the issues above would be addressed if I move this change into `ra_server_proc` and add new state(s). This comes with its own tradeoffs, e.g. code duplication with the follower state. I am prototyping this on a separate branch.

----

Updated transition diagram:

```
      ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─          ┌──────────────────────────┐
─────▷  voter => {no, Condition} │────────▷│ voter => {no, permanent} │
      └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─          └──────────────────────────┘
                    │                                                  
                    │                                                  
                    ▽                    ▷ $ra_join (now can be used to update voter status)
      ┌──────────────────────────┐       ▶ initial_cluster_members                   
─────▶│       voter => yes       │
      └──────────────────────────┘                                     
```

src/ra.hrl
```diff
@@ -44,12 +44,18 @@
                           suspended |
                           disconnected.
 
+-type ra_voter_status() :: yes | {no, ra_nonvoter_reason()}.
```
kjnilsson, line 47: should it be `voter` / `non_voter` instead of `yes` / `no`?

src/ra.hrl
```diff
@@ -44,12 +44,18 @@
...
+-type ra_nonvoter_reason() :: #{target := ra_index()}.
+
 -type ra_peer_state() :: #{next_index := non_neg_integer(),
                            match_index := non_neg_integer(),
                            query_index := non_neg_integer(),
                            % the commit index last sent
                            % used for evaluating pipeline status
                            commit_index_sent := non_neg_integer(),
+                           %% whether the peer is part of the consensus
+                           voter := ra_voter_status(),
```
kjnilsson, line 58: `:=` makes this field required in the map, is it required always?

src/ra_server.erl
```diff
@@ -1104,8 +1111,16 @@ handle_follower({ra_log_event, Evt}, State = #{log := Log0}) ->
     % simply forward all other events to ra_log
     {Log, Effects} = ra_log:handle_event(Evt, Log0),
     {follower, State#{log => Log}, Effects};
+handle_follower(#pre_vote_rpc{},
+                #{cfg := #cfg{log_id = LogId}, voter := {no, _} = Voter} = State) ->
+    ?DEBUG("~w: follower ignored pre_vote_rpc, non-voter: ~p", [LogId, Voter]),
```
kjnilsson, line 1116: use `~s` for the LogId and `~p0` for the Voter to ensure we avoid multi line logs.

src/ra_server.erl
```diff
@@ -2451,16 +2479,34 @@ append_log_leader({CmdTag, _, _, _},
...
     case OldCluster of
         #{JoiningNode := _} ->
             % already a member do nothing
             % TODO: reply? If we don't reply the caller may block until timeout
             {not_appended, already_member, State};
         _ ->
-            Cluster = OldCluster#{JoiningNode => new_peer()},
-            append_cluster_change(Cluster, From, ReplyMode, State)
+            append_log_leader({'$ra_join', From,
+                            #{node => JoiningNode, voter => new_nonvoter(State)},
```
kjnilsson, line 2508: indentation

src/ra_server.erl
```diff
@@ -2804,6 +2855,49 @@ meta_name(#cfg{system_config = #{names := #{log_meta := Name}}}) ->
...
+%%% Voter status helpers
+%%% ====================
+
+-spec new_nonvoter(ra_server_state()) -> ra_voter_status().
+
+new_nonvoter(#{commit_index := Target} = _State) ->
+    {no, #{target => Target}}.
+
+-spec maybe_promote_voter(ra_server_id(), ra_server_state(), effects()) -> effects().
+
```
kjnilsson, line 2869: remove blank line

src/ra_server.erl
```diff
@@ -2804,6 +2855,49 @@ meta_name(#cfg{system_config = #{names := #{log_meta := Name}}}) ->
...
+    end.
+
+update_voter_status({no, #{target := Target}}, MI)
+  when MI >= Target ->
+    yes;
+update_voter_status(Permanent, _) ->
+    Permanent.
+
+-spec voter_status(ra_server_id(), ra_cluster()) -> ra_voter_status().
+
```
kjnilsson, line 2892: remove blankline

src/ra_server.erl
```diff
@@ -1370,12 +1390,18 @@ overview(#{cfg := #cfg{effective_machine_module = MacMod} = Cfg,
...
                     voted_for,
                     cluster_change_permitted,
                     cluster_index_term,
                     query_index
                    ], State),
-    O = maps:merge(O0, cfg_to_map(Cfg)),
+    % implicit voter status for initial leaders
+    O1 = case O0 of
+             #{voter := _} -> O0;
+             _             -> O0#{voter => yes}
```
kjnilsson, line 1402: remove redundant whitespace after `_` - if you want to line up the expressions use a newline.

src/ra_server.erl
```diff
@@ -2451,16 +2479,34 @@ append_log_leader({CmdTag, _, _, _},
...
+            Cluster = OldCluster#{JoiningNode => Peer#{voter => Voter}},
+            append_cluster_change(Cluster, From, ReplyMode, State);
+        _ ->
+            % Insert new member.
+            Cluster = OldCluster#{JoiningNode => new_peer_with(#{voter => Voter})},
+            append_cluster_change(Cluster, From, ReplyMode, State)
+    end;
 append_log_leader({'$ra_join', From, JoiningNode, ReplyMode},
                   State = #{cluster := OldCluster}) ->
+    % Legacy $ra_join, join as future voter iff no such member in the cluster.
```
kjnilsson, line 2500: "join as a non-voter"?

src/ra_server.erl
```diff
@@ -2804,6 +2855,49 @@ meta_name(#cfg{system_config = #{names := #{log_meta := Name}}}) ->
     Name;
 meta_name(#{names := #{log_meta := Name}}) ->
     Name.
+
+%%% ====================
+%%% Voter status helpers
+%%% ====================
+
+-spec new_nonvoter(ra_server_state()) -> ra_voter_status().
+
```
kjnilsson, line 2864: remove blank line

src/ra_server.erl
```diff
@@ -2451,16 +2479,34 @@ append_log_leader({CmdTag, _, _, _},
...
+append_log_leader({'$ra_join', From, #{node := JoiningNode, voter := Voter}, ReplyMode},
+                  State = #{cluster := OldCluster}) ->
+    case OldCluster of
+        #{JoiningNode := #{voter := Voter}} ->
+            % already a member do nothing
+            % TODO: reply? If we don't reply the caller may block until timeout
+            {not_appended, already_member, State};
+        #{JoiningNode := Peer} ->
+            % Update member status.
+            Cluster = OldCluster#{JoiningNode => Peer#{voter => Voter}},
```
kjnilsson, line 2491: Does this mean that voter demotion is also possible?

kjnilsson (changes requested):
Looks good, some formatting and nitpicks only.

src/ra_server.erl
```diff
@@ -2803,6 +2850,62 @@ meta_name(#cfg{system_config = #{names := #{log_meta := Name}}}) ->
...
+-spec required_quorum(ra_cluster()) -> pos_integer().
+
+required_quorum(Cluster) ->
+    Voters = count_voters(Cluster),
+    trunc(Voters / 2) + 1.
+
+count_voters(Cluster) ->
+    maps:fold(
+      fun (_, #{voter := {no, _}}, Count) -> Count;
+          (_, _, Count)                   -> Count + 1
```
kjnilsson, line 2905: Use the same formatting in the rest of the code base (i.e. remove redunantant indentation before `->` and favour putting the expression that follows on a new line. :)

src/ra_server.erl
```diff
@@ -2803,6 +2850,62 @@ meta_name(#cfg{system_config = #{names := #{log_meta := Name}}}) ->
...
+voter_status(PeerId, Cluster) ->
+    case maps:get(PeerId, Cluster, undefined) of
+        undefined ->
+            {no, undefined};
+        Peer ->
+            maps:get(voter, Peer, yes)
+    end.
+
+-spec required_quorum(ra_cluster()) -> pos_integer().
+
```
kjnilsson, line 2897: Pls remove this blank line.

test/ra_SUITE.erl
```diff
@@ -641,6 +642,9 @@ add_member(Config) ->
     C = {ra_server:name(Name, "3"), node()},
     ok = ra:start_server(default, Name, C, add_machine(), Cluster),
     {ok, _, _Leader} = ra:add_member(Leader, C),
+    {ok, Members, _From} = ra:members(A),
+    Ms = [ ra:member_overview(P) || {_, _} = P <- Members ],
```
kjnilsson, line 646: spaces not needed after `[` and before `]`

test/ra_SUITE.erl
```diff
@@ -1126,6 +1169,21 @@ nth_server_name(Config, N) when is_integer(N) ->
 add_machine() ->
     {module, ?MODULE, #{}}.
 
+overviews(Node) ->
+    {ok, Members, _From} = ra:members(Node),
+    [ ra:member_overview(P) || {_, _} = P <- Members ].
+
+voters({ok, #{cluster := Peers}, _} = _Overview) ->
+    [ Id || {Id, Status} <- maps:to_list(Peers), maps:get(voter, Status) =:= yes ].
```
kjnilsson, line 1177: spaces around brackets not needed

test/ra_SUITE.erl
```diff
@@ -1126,6 +1169,21 @@ nth_server_name(Config, N) when is_integer(N) ->
...
+
+is_voter({ok, #{voter := yes}, _} = _Overview) ->
+    true;
+is_voter({ok, _, _} = _Overview) ->
+    false.
+
+nonvoters({ok, #{cluster := Peers}, _} = _Overview) ->
+    [ Id || {Id, Status} <- maps:to_list(Peers), maps:get(voter, Status) /= yes ].
```
kjnilsson, line 1185: spaces around brackets

src/ra.hrl
illotum: I'm trying to avoid stuttering a-la "voter => voter". And naming is hard ofc.

How about then changing the map key as well? Any suggestions?

src/ra.hrl
illotum: I have no strong opinion. My only reasoning was that it is easier to deconstruct args when the field is required.

src/ra_server.erl
illotum: Correct! Nothing in Ra does that yet, but my reasoning is that library users will appreciate the capability.

src/ra_server.erl
```diff
@@ -2818,6 +2889,99 @@ meta_name(#cfg{system_config = #{names := #{log_meta := Name}}}) ->
...
+        Self -> New;
+        _ -> Current
+    end.
+
+-spec maybe_promote_peer(ra_server_id(), ra_server_state(), effects()) -> effects().
+maybe_promote_peer(PeerID, #{cluster := Cluster}, Effects) ->
+    % Unknown peer handled in the caller.
+    #{PeerID := #{match_index := MI,
+                  voter_status := OldStatus}} = Cluster,
+    case update_voter_status(OldStatus, MI) of
```
kjnilsson, line 2953: I think it may be easier to read if the logic form this single use function was inlined here as a pattern match

src/ra_server.erl
```diff
@@ -195,6 +197,7 @@
                               max_pipeline_count => non_neg_integer(),
                               ra_event_formatter => {module(), atom(), [term()]},
                               counter => counters:counters_ref(),
+                              init_non_voter => ra_nvid(),
```
kjnilsson, line 200: could we name this `non_voter_id`?

src/ra_server.erl
```diff
@@ -297,13 +300,17 @@ init(#{id := Id,
         case ra_log:recover_snapshot(Log0) of
             undefined ->
                 InitialMachineState = ra_machine:init(Machine, Name),
-                {0, make_cluster(Id, InitialNodes),
+                Clu = make_cluster(Id, InitialNodes,
```
kjnilsson, line 303: is it possible to set initial nodes which are non voters? if not we shouldn't need to take the initial voter status from config here.

src/ra_server.erl
```diff
@@ -2818,6 +2889,99 @@ meta_name(#cfg{system_config = #{names := #{log_meta := Name}}}) ->
...
+%%% Voter status helpers
+%%% ====================
+
+-spec ensure_promotion_target(ra_voter_status(), ra_server_state()) ->
+    {ok, ra_voter_status()} | {error, term()}.
+ensure_promotion_target({voter, Reason}, _) ->
+    {ok, {voter, Reason}};
+ensure_promotion_target({nonvoter, #{target := _, nvid := _} = Reason}, _) ->
+    {ok, {nonvoter, Reason}};
+ensure_promotion_target({nonvoter, #{nvid := _} = Reason}, #{commit_index := CI}) ->
```
kjnilsson, line 2908: I'm thinking using `ra_log:next_index/1` to get the promotion target would be more "correct", it should in this case be the same log index as the non voter was added at.

kjnilsson: Also this line could do with a newline somewhere, It goes over the 80 char mark.

src/ra.erl
```diff
@@ -716,6 +719,13 @@ new_uid(Source) when is_binary(Source) ->
     Prefix = ra_lib:derive_safe_string(Source, 6),
     ra_lib:make_uid(string:uppercase(Prefix)).
 
+%% @doc generates a random uid using timestamp for the first
+%% 6 characters.
+%% @end
+new_nvid() ->
```
kjnilsson, line 725: Do we need a prefix at all?

src/ra_server_proc.erl
```diff
@@ -1717,9 +1726,11 @@ handle_tick_metrics(State) ->
     _ = ets:insert(ra_metrics, Metrics),
     State.
 
-can_execute_locally(RaftState, TargetNode, State) ->
+can_execute_locally(RaftState, TargetNode,
+                    #state{server_state = ServerState} = State) ->
+    {Voter, _} = ra_server:voter_status(ServerState),
     case RaftState of
-        follower ->
+        follower when Voter == voter ->
```
kjnilsson, line 1733: The next clause (`leader when TargetNode =/= node()`) will probably need to take voting status into account else the leader may not execute the effect instead of the non voter.

kjnilsson: coordination_SUITE has some local tests so that may be a good place to put a test for this.

kjnilsson (changes requested):
Looking good. I have a few smaller suggestions.

I will start testing this PR with RabbitMQ

src/ra_server.erl
illotum: I had the key exactly at that, but one consideration made me change it: unlike UID this one can (hypothetically) be changed with follow-up cluster changes. How about `initial_non_voter_id`?

src/ra_server.erl
illotum: This value is used to add the node itself during a cluster join, L2262. Since there is a top-level voter_status, seeding the one within cluster is optional, but I chose to keep it consistent. LMK if you prefer it pure.

src/ra.erl
illotum: It is there mostly to compensate for non-crypto source in the random part.

But related: I've been on and off the idea of re-using UID for non-voter identity. I chose a separate ID for clarity, but there is appeal in re-using what we already have. What do you think?

michaelklishin: I don't think a node's identity changes when its status become a non-voter, so I'd say the same ID can be used?

src/ra_server_proc.erl
illotum: Done via a new local query, tests included 👍

illotum:
Rebased and pushed the uid change. It is one of those small things that cascade widely, but a sum PR diff became saner IMO. Notably, local state and ETS now only hold `non_voter :: boolean`, and that's what various reporting tools expose.

Tomorrow I will deal with dialyzer.

src/ra.erl
```diff
@@ -1132,13 +1133,16 @@ key_metrics({Name, N} = ServerId) when N == node() ->
                end,
     case whereis(Name) of
         undefined ->
-            Counters#{state => noproc};
+            Counters#{state => noproc,
+                      non_voter => noproc};
```
kjnilsson, line 1137: I think this should be `unknown` rather than `noproc` - I need to think of some way to report the leader's view of voter status in the case where some members are not reachable. It doesn't affect this PR though.

src/ra.hrl
```diff
@@ -39,17 +39,33 @@
...
 -type ra_server_id() :: {Name :: atom(), Node :: node()}.
 
+%% Specifies server configuration for a new cluster member.
+%% Subset of  ra_server:ra_server_config().
+%% Both `ra:add_member` and `ra:start_server` must be called with the same values.
+-type ra_new_server() :: #{id := ra_server_id(),
+
+                           %% If set, server will start as non-voter until later promoted by the
+                           %% leader.
+                           non_voter => boolean(),
```
kjnilsson, line 49: I'm not a big fan of booleans I have to admit and I think we should consider how we might want to configure a new member which isn't automatically promoted to a full voter. In this case we have 3 joining member types: `voter | promotable_non_voter | permanent_non_voter`. 

So perhaps something like `membership` => `voter | promotable | non_voter`

src/ra.hrl
```diff
@@ -39,17 +39,33 @@
...
+                             target => ra_index()}.
+
 -type ra_peer_state() :: #{next_index := non_neg_integer(),
                            match_index := non_neg_integer(),
                            query_index := non_neg_integer(),
                            % the commit index last sent
                            % used for evaluating pipeline status
                            commit_index_sent := non_neg_integer(),
+                           %% whether the peer is part of the consensus
+                           voter_status := ra_voter_status(),
```
kjnilsson, line 68: is `voters_status` a mandatory key?

src/ra.hrl
```diff
@@ -139,6 +155,7 @@
 -type snapshot_meta() :: #{index := ra_index(),
                            term := ra_term(),
                            cluster := ra_cluster_servers(),
+                           cluster_state => ra_cluster(),  %% TODO replace `cluster`
```
kjnilsson, line 158: would we really persist the whole `ra_cluster()` map in the snapshot? Only the voter_status is really required the rest of the fields are discovered during recovery etc.

kjnilsson (changes requested):
I have done some testing in RabbitMQ with this Pr and so far it all looks good now.

Some questions and suggestions included which we need to address.

src/ra.hrl
```diff
@@ -43,18 +43,19 @@
 %% Subset of  ra_server:ra_server_config().
 %% Both `ra:add_member` and `ra:start_server` must be called with the same values.
 -type ra_new_server() :: #{id := ra_server_id(),
-
-                           %% If set, server will start as non-voter until later promoted by the
-                           %% leader.
-                           non_voter => boolean(),
+                           % Defaults to `voter` is absent.
```
michaelklishin, line 46: I think this was meant to say "if absent"

src/ra_server.erl
```diff
@@ -6,6 +6,8 @@
 %%
 -module(ra_server).
 
+-feature(maybe_expr, enable).
```
kjnilsson, line 9: Can we refactor all code not to use `maybe` expressions? Currently it has broken both the OTP 24 _and_ 25 tests. We can probably fix the 25 tests but it would be nice to still stay 24 compatible a little longer (as it is a library).

```diff --git a/src/ra.erl b/src/ra.erl
index 159628f..8202a7b 100644
--- a/src/ra.erl
+++ b/src/ra.erl
@@ -415,7 +415,7 @@ start_cluster(System, [#{cluster_name := ClusterName} | _] = ServerConfigs,
                     ok  -> true;
                     Err ->
                         ?ERR("ra: failed to start a server ~w, error: ~p",
-                              [C, Err]),
+                              [maps:get(id, C), Err]),
                         false
                 end
             end, ServerConfigs),
diff --git a/src/ra_server.erl b/src/ra_server.erl
index a16fe40..f4f60a0 100644
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -6,8 +6,6 @@
 %%
 -module(ra_server).
 
--feature(maybe_expr, enable).
-
 -include("ra.hrl").
 -include("ra_server.hrl").
 
@@ -2947,22 +2945,27 @@ get_membership(#{cfg := #cfg{id = Id, uid = UId}, cluster := Cluster} = State) -
     Default = maps:get(membership, State, voter),
     get_membership(Cluster, Id, UId, Default).
 
--spec maybe_promote_peer(ra_server_id(), ra_server_state(), effects()) -> effects().
+-spec maybe_promote_peer(ra_server_id(), ra_server_state(), effects()) ->
+    effects().
 maybe_promote_peer(PeerId, #{cluster := Cluster}, Effects) ->
-    maybe
+    case Cluster of
         #{PeerId := #{match_index := MI,
-                      voter_status := VoterStatus}} ?= Cluster,
-        #{membership := promotable, target := Target} ?= VoterStatus,
-        true ?= MI >= Target,
-        E = {next_event,
-          {command, {'$ra_join',
-                    #{ts => os:system_time(millisecond)},
-                    #{id => PeerId,
-                      voter_status => VoterStatus#{membership => voter}},
-                    noreply}}},
-        [E | Effects]
-    else
-        _ -> Effects
+                      voter_status := VoterStatus}}  ->
+            case VoterStatus of
+                #{membership := promotable, target := Target} when
+                      MI >= Target ->
+                    [{next_event,
+                      {command, {'$ra_join',
+                                 #{ts => os:system_time(millisecond)},
+                                 #{id => PeerId,
+                                   voter_status => VoterStatus#{membership => voter}},
+                                 noreply}}},
+                     Effects];
+                _ ->
+                    Effects
+            end;
+        _ ->
+            Effects
     end.
 
 -spec required_quorum(ra_cluster()) -> pos_integer().
diff --git a/src/ra_server_proc.erl b/src/ra_server_proc.erl
index 16326ad..c985aa4 100644
--- a/src/ra_server_proc.erl
+++ b/src/ra_server_proc.erl
@@ -7,8 +7,6 @@
 %% @hidden
 -module(ra_server_proc).
 
--feature(maybe_expr, enable).
-
 -behaviour(gen_statem).
 
 -compile({inline, [handle_raft_state/3]}).
diff --git a/test/coordination_SUITE.erl b/test/coordination_SUITE.erl
index 4e8f065..c6a45df 100644
--- a/test/coordination_SUITE.erl
+++ b/test/coordination_SUITE.erl
@@ -56,6 +56,10 @@ groups() ->
 
 init_per_suite(Config) ->
     %% as we're not starting the ra application and we want the logs
+    LogFile = filename:join(?config(priv_dir, Config), "ra.log"),
+    LogCfg = #{config => #{file => LogFile}},
+    logger:add_handler(test_handler, logger_std_h, LogCfg),
+    logger:set_primary_config(level, debug),
     ra_env:configure_logger(logger),
     Config.
 
@@ -112,6 +116,7 @@ start_stop_restart_delete_on_remote(Config) ->
     ok.
 
 start_cluster(Config) ->
+    logger:set_primary_config(level, debug),
     PrivDir = ?config(data_dir, Config),
     ClusterName = ?config(cluster_name, Config),
     NodeIds = [{ClusterName, start_follower(N, PrivDir)} || N <- [s1,s2,s3,s4,s5]],
@@ -722,7 +727,7 @@ start_follower(N, PrivDir) ->
     ct:pal("starting secondary node with ~ts on host ~ts for node ~ts", [Pa, Host, node()]),
     {ok, S} = slave:start_link(Host, N, Pa),
     ok = ct_rpc:call(S, ?MODULE, node_setup, [PrivDir]),
-    ok = erpc:call(S, ra, start, []),
+    ok = ct_rpc:call(S, ra, start, []),
     ok = ct_rpc:call(S, logger, set_primary_config,
                      [level, all]),
     S.
```

PR #388, opened 2023-07-17
Handle shutdown exits in gen_statem:call/3

the-mikedavis:
## Proposed Changes

`gen_statem:call/3` can fail if the server shuts down before replying to the call, exiting with `{shutdown, Location}`. We can handle this exit like the other exits in the `gen_statem_safe_call/3`. I think it makes sense to return `{error, noproc}` in this case since it's effectively the same as the server not being alive, but I think an `{error, shutdown}` return could be ok as well.

## Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Cosmetics

## Checklist

### DIFF
--- a/src/ra.hrl
+++ b/src/ra.hrl
@@ -65,7 +65,7 @@
 
 -type consistent_query_ref() :: {From :: term(), Query :: ra:query_fun(), ConmmitIndex :: ra_index()}.
 
--type safe_call_ret(T) :: timeout | {error, noproc | nodedown} | T.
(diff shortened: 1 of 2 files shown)


michaelklishin:
I like the shutdown one better. `noproc` more or less never means an intended return value, or at least some people automatically expect it to indicate an *unhandled* error. WDYT?

the-mikedavis:
Yeah that sounds good to me. `noproc` is also a little different than `shutdown` here since the server did actually receive the command with `shutdown`, so let's go with `shutdown`.

michaelklishin (approved):

PR #391, opened 2023-08-22
Use seshat from bazel-central-registry

pjk25:
making it a native bazel_dep


michaelklishin:
#386 has been merged :)

PR #401, opened 2023-11-01
Quote `maybe` atoms

sile:
## Proposed Changes

Basically, `ra` codebase quotes `maybe` atoms (to support the maybe feature in the future I guess). 
However there are some `maybe` atoms that are not quoted, and `ra` can't be compiled if the maybe feature is enabled.
This PR fixes this problem.
Of course, this is not a critical problem. So, feel free to close this PR if it seems not needed.

## Types of Changes

- [x] Cosmetics (whitespace, appearance)

## Checklist

- [x] Any dependent changes have been merged and published in related repositories

## Further Comments

considered, etc.

### DIFF
--- a/src/ra_log.erl
+++ b/src/ra_log.erl
@@ -546,7 +546,7 @@ next_index(#?MODULE{last_index = LastIdx}) ->
     LastIdx + 1.
 
 -spec fetch(ra_index(), state()) ->
-    {maybe(log_entry()), state()}.
+    {'maybe'(log_entry()), state()}.
 fetch(Idx, State0) ->
     case fold(Idx, Idx, fun(E, Acc) -> [E | Acc] end, [], State0) of
         {[], State} ->
--- a/src/ra_log_segment.erl
+++ b/src/ra_log_segment.erl
@@ -59,7 +59,7 @@
          data_start :: pos_integer(),
          data_offset :: pos_integer(),
          data_write_offset :: pos_integer(),
-         index = undefined :: maybe(ra_segment_index()),
+         index = undefined :: 'maybe'(ra_segment_index()),
          range :: 'maybe'({ra_index(), ra_index()}),
          pending_data = [] :: iodata(),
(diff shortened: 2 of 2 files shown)


michaelklishin:
`maybe` in this case is a type name. It's a bit unexpected that the same quoting rules would apply to type specs since there cannot be any conflict with the `maybe` keyword.

Also, Ra is compiled and passes all tests on CI on Erlang 26 (where `maybe` is supposedly always enabled). But Make-based tests fail with this PR.

What am I missing?

michaelklishin:
Maybe (pun intended) we should sidestep the issue entirely :)  https://github.com/rabbitmq/ra/pull/402

sile:
Thank you for your comments.

Seeing [the OTP 26 highlight](https://www.erlang.org/blog/otp-26-highlights/#no-need-to-enable-feature-maybe-in-the-runtime-system), the runtime's maybe feature is enabled by default, but the compiler's is not.
If we enabled the compiler's maybe feature, ra cannot be compiled as follows:
```console
// Erlang/OTP version
$ erl -boot start_clean -noshell -eval 'io:format(erlang:system_info(system_version)).' -s init stop
Erlang/OTP 26 [erts-14.1] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]

// Clone ra
$ git clone https://github.com/rabbitmq/ra.git 
$ cd ra/

// Compile ra_log.erl without and with the maybe_expr compile feature 

// without (default) => OK
$ erlc src/ra_log.erl

// with => NG
$ erlc -enable-feature maybe_expr src/ra_log.erl
src/ra_log.erl:549:6: syntax error before: 'maybe'
%  549|     {maybe(log_entry()), state()}.
%     |      ^
```

sile:
> Maybe (pun intended) we should sidestep the issue entirely :) https://github.com/rabbitmq/ra/pull/402 

I see. The approach of #402 is perfect for me :)
Thank you!

PR #407, opened 2024-01-05
Fixes to ra_leaderboard module

kjnilsson:
Ensure that the members are updated when the membership
is modifiied.

Always delete leaderboard record in ra server terminate callback
to favour accuracy of leadership information over availability.

The idea is that when a record is not found in the leaderboard
the caller falls back to some other, potentially slower, means
of leader discovery.


dumbbell (approved):
The test of rabbitmq/khepri#240 (to use `ra_leaderboard:lookup_members/1`)  works with this patch works!

It failed with Ra 2.7.1.

kjnilsson:
> The test of [rabbitmq/khepri#240](https://github.com/rabbitmq/khepri/pull/240) (to use `ra_leaderboard:lookup_members/1`) works with this patch!
> 
> It failed with Ra 2.7.1.

nice, thanks for testing, I'll release a new Ra patch

PR #409, opened 2024-01-12
rebar: Sync aten version pinning with `Makefile`

dumbbell:
## Why

It was bumped to 0.6.0 in commit 3503c79b04f694c6c33ece8feb7fe1c2036cc3f4. `rebar.config` was not updated at the same time.


dumbbell:
FTR, Khepri is using Rebar. This problem was detected because after updating the Ra pinning in Khepri to 2.7.2, aten was left at 0.5.8.

PR #411, opened 2024-01-23
add API to list voters

illotum:
## Proposed Changes

This is a small addition, high-level API to list voters of a Ra cluster. To be used wherever an active quorum list is needed instead of ra:members.

## Types of Changes

## Checklist
- [x] I have added necessary documentation (if appropriate)

### DIFF
--- a/src/ra.erl
+++ b/src/ra.erl
@@ -23,6 +23,8 @@
          %% queries
          members/1,
          members/2,
+         voters/1,
+         voters/2,
          initial_members/1,
          initial_members/2,
          local_query/2,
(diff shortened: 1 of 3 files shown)


michaelklishin (approved):

michaelklishin:
@illotum our team would like to ask for another API extension. This one is fine for where it will be used in RabbitMQ, so I guess we can keep it.

The function we'd like to see is a version of `ra:members/1` that would return a map with more detailed view of cluster members.

`ra:members_with_details/1` and `ra:members_ext/1` are two named proposed. WDYT?

illotum:
For sure! I usually lean towards minimal changes, but this covers all my needs and more.

kjnilsson:
> For sure! I usually lean towards minimal changes, but the extended API covers all my needs and more.

I think with all the peer info we can probably make better external decision on stuff like member changes without further probing. I have on a couple of occasions wished I could get this information in an easier way. `ra:voters/1` is very specific the  `ra` module API is growing quite large and having a more generic function avoids us having to add yet another function in the future.

illotum:
Following on the rabbitmq-server/pull/10394 thread, lets revert this merge then and I'll go ahead with a proper member_info.

michaelklishin:
Done.

PR #413, opened 2024-01-24
add members_info state query and API wrapper

illotum:
## Proposed Changes

Adds state query and API to return leader's cluster view.

## Types of Changes

## Checklist


illotum:
Opening for review. 

@kjnilsson what if we called it `cluster_info` instead? Considering the return type and internal representation, it feels a little more intuitive.

Second question, should I send zeroes instead of no values for the ease of matching? This mainly concerns the corner case of a local query to non-leader.

michaelklishin:
@illotum `cluster_info` is not a bad name but note that a cluster status is (or at least can be) more than the status of its members. `members_info` is more specific in that sense.

src/ra_server_proc.erl
```diff
@@ -1503,13 +1505,7 @@ follower_leader_change(Old, #state{pending_commands = Pending,
...
     case node() of
         Node -> ok;
-        _ ->
-            case aten:register(Node) of
-                ignore ->
-                    ok;
-                Res ->
-                    Res
-            end
+        _ -> aten:register(Node)
```
michaelklishin, line 1508: This is not a direct equivalent (maybe this does not matter). In the original version a value of `ignore` returned by Aten would be returned as `ok` by Ra. Now that's not the case, is there a reason for that?

illotum: Dialyze mostly! `aten:register` returns only `ok`.

illotum:
Added a commit to extract membership from `voter_status` key. I'm uncertain if this is better than full voter_status, but clearly it is the main value of that entry. I can think of a situation where `target` may be valuable, so happy to revert.

michaelklishin (approved):

src/ra_server_proc.erl
```diff
@@ -1555,6 +1551,51 @@ do_state_query(voters, #{cluster := Cluster}) ->
...
+    maps:map(fun(Id, Peer) ->
+                 case {Id, Peer} of
+                     {Self, Peer = #{voter_status := _}} ->
+                         %% For completeness sake, preserve `target`
+                         %% of once promoted leader.
+                         #{next_index => CI+1,
+                          match_index => CI,
+                          query_index => QI,
+                          status => normal,
+                          voter_status => Peer#{membership => Membership}};
```
kjnilsson, line 1566: Sould this be the voter status map instead of the full `Peer`?

src/ra_server_proc.erl
kjnilsson: I think you have an old aten version, see https://github.com/rabbitmq/aten/blob/main/src/aten.erl#L18

kjnilsson (changes requested):
Couple of smaller changes needed but looks good.

Also would you mind running an indentation pass on the new code in ra_server_proc. It's all a bit wonky :)

kjnilsson:
> Opening for review.
> 
> @kjnilsson what if we called it `cluster_info` instead? Considering the return type and internal representation, it feels a little more intuitive.


I think the link to `ra:members/1` works in terms of it's name so let's keep it as is. 

> 
> Second question, should I send zeroes instead of no values for the ease of matching? This mainly concerns the corner case of a local query to non-leader. On the flip side, I fear sending anything invalid will lead to user confusion.

I think what we're hitting here is the fact that this query makes less sense as a local query as really you want to issue it against the leader. I suggest when used as a local query we only return infos that make sense to return, e.g. the a=`voter_status`. Alt. we always make this query a leader query. Happy to consider either way.

src/ra_server_proc.erl
illotum: That's what I get for working late night!

illotum:
I'm leaning towards returning _something_. There are cases where one may want to investigate follower's view of the cluster, however limited that is.

michaelklishin:
In the context of RabbitMQ's needs of determining if a node hosts any "catching up" (non-voter) replicas, either can work but a leader query makes more sense to me.

michaelklishin (approved):

Issue #141, opened 2019-11-12
Ra checkpoints


kjnilsson:
Some state machines that deliberately leave a long log in place (e.g. RabbitMQ quorum queues may do this as messages on queue is increasing) may want to create checkpoints which effectively are snapshots that do not truncate the log.

This  will help speeding up the recovery phase which can be very slow if the log is very large as the server can use the latest checkpoint rather than the snapshot as a recovery starting point. Checkpoints could even be written during an orderly shutdown of the ra server.

It can also help reduce memory overhead of RabbitMQ quorum queues as they currently keep something akin to checkpoints in memory.

Machines would use a new effect: `{checkpoint, ra_index(), machine_state()}` to emit new checkpoints. A server could have many checkpoints but it is likely that we'd need some upper limit. Once the upper limit is reached Ra would "thin" the list of checkpoints such that the oldest and newest checkpoints are always in retained but checkpoints between them would become further and further apart.

Checkpoints can be promoted to snapshots using a new `{release_cursor, ra_index()}` effect which will promote the checkpoint with the highest index that is lower or equal to the release_cursor index to a snapshot and delete all checkpoints up to and including the promoted checkpoint.

Any checkpoints with an index lower than the current snapshot should be removed.

Ra should avoid writing checkpoints if the last checkpoint was not written too many indexes ago. (e.g. 4096) to avoid a proliferation of checkpoint work.

Checkpoints are kept in the same structure as snapshots in a directory called `checkpoints` that is adjacent to the `snapshots` directory.

PR #415, opened 2024-01-29
Checkpoints

the-mikedavis:
As described in #141, we add a new effect that machines may emit:

```erl
{checkpoint, CheckpointIndex, MachineState}
```

This suggests that Ra should add a checkpoint. Checkpoints are essentially the same as snapshots except that they don't trigger log truncation. They reuse the `ra_snapshot` behavior, so they are exactly the same as snapshots on disk. They can later be promoted to actual snapshots by emitting another new effect:

```erl
{release_cursor, ReleaseCursorIndex}
```

which suggests that Ra should find the checkpoint with the highest index lower than or equal to `ReleaseCursorIndex` and rename the checkpoint file so that it moves to `snapshots/` from `checkpoints/`. Any checkpoints lower than that index are then deleted, and log can be truncated up to that index. There is a configurable maximum number of checkpoints allowed. Adding more checkpoints after that maximum will trigger a "thinning out" where checkpoints are deleted randomly from the middle of the checkpoint list.

`rabbit_fifo` currently has an ad-hoc checkpointing system where it queues snapshot effects (`{release_cursor, RaftIndex, DehydratedState}`) regularly and emits those effects when it moves up the release cursor. The advantage of building this into Ra is that we can store the checkpoints on disk, so we can use them for machine recovery and reduce memory consumption (albeit at the cost of disk / IO).

This is useful for machines like `rabbit_fifo` that need to keep the log around on disk for a potentially long time (for use with the `{log, Idxs, Fun, Opts}` effect). If we take checkpoints regularly then recovery becomes constant-time rather than linear on the number of messages in the queue. Testing this locally on my machine against the server with a QQ containing 5 million messages[^1] gives a recovery time of 10ms while `main` takes around 24s.

Closes https://github.com/rabbitmq/ra/issues/141

[^1]: use the [`md-ra-checkpoints`](https://github.com/rabbitmq/rabbitmq-server/tree/md-ra-checkpoints) branch, fill the queue with `perf-test -x 1 -y 0 -qq -u qq -c 3000 -C 5000000`, restart the broker and grep the log file for "recovery of state machine version"

### DIFF
--- a/src/ra_log.erl
+++ b/src/ra_log.erl
@@ -505,37 +505,55 @@ handle_event({segments, Tid, NewSegs},
                         end),
             {State, log_update_effects(Readers, Pid, State)}
     end;
-handle_event({snapshot_written, {SnapIdx, _} = Snap, _SnapKind},
+handle_event({snapshot_written, {SnapIdx, _} = Snap, SnapKind},
              #?MODULE{cfg = Cfg,
                       first_index = FstIdx,
                       snapshot_state = SnapState0} = State0)
 %% only update snapshot if it is newer than the last snapshot
   when SnapIdx >= FstIdx ->
-    % delete any segments outside of first_index
-    {State, Effects0} = delete_segments(SnapIdx, State0),
-    SnapState = ra_snapshot:complete_snapshot(Snap, SnapState0),
-    put_counter(Cfg, ?C_RA_SVR_METRIC_SNAPSHOT_INDEX, SnapIdx),
-    %% delete old snapshot files
-    %% This is done as an effect
-    %% so that if an old snapshot is still being replicated
-    %% the cleanup can be delayed until it is safe
-    Effects = [{delete_snapshot,
-                ra_snapshot:directory(SnapState),
-                ra_snapshot:current(SnapState0)} | Effects0],
-    %% do not set last written index here as the snapshot may
-    %% be for a past index
-    {State#?MODULE{first_index = SnapIdx + 1,
-                   snapshot_state = SnapState}, Effects};
-handle_event({snapshot_written, {Idx, Term} = Snap, _SnapKind},
+    SnapState1 = ra_snapshot:complete_snapshot(Snap, SnapKind, SnapState0),
+    case SnapKind of
+        snapshot ->
+            put_counter(Cfg, ?C_RA_SVR_METRIC_SNAPSHOT_INDEX, SnapIdx),
+            % delete any segments outside of first_index
+            {State, Effects0} = delete_segments(SnapIdx, State0),
+            %% Delete old snapshot files. This is done as an effect
+            %% so that if an old snapshot is still being replicated
+            %% the cleanup can be delayed until it is safe.
+            DeleteCurrentSnap = {delete_snapshot,
+                                 ra_snapshot:directory(SnapState1, snapshot),
+                                 ra_snapshot:current(SnapState0)},
+            %% Also delete any checkpoints older than this snapshot.
+            {SnapState, Checkpoints} =
+                ra_snapshot:take_older_checkpoints(SnapIdx, SnapState1),
+            CPEffects = [{delete_snapshot,
+                          ra_snapshot:directory(SnapState, checkpoint),
+                          Checkpoint} || Checkpoint <- Checkpoints],
+            Effects = [DeleteCurrentSnap | CPEffects] ++ Effects0,
+            %% do not set last written index here as the snapshot may
+            %% be for a past index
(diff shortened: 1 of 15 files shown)


src/ra_server.erl
```diff
@@ -1848,6 +1850,22 @@ update_release_cursor(Index, MacState,
                                                   MacState, Log0),
     {State#{log => Log}, Effects}.
 
+-spec checkpoint(ra_index(), term(), ra_server_state()) ->
+      {ra_server_state(), effects()}.
+
+checkpoint(Index, MacState,
+               State = #{log := Log0, cluster := Cluster}) ->
```
kjnilsson, line 1857: indentation

src/ra_server.erl
```diff
@@ -1848,6 +1850,22 @@ update_release_cursor(Index, MacState,
                                                   MacState, Log0),
     {State#{log => Log}, Effects}.
 
+-spec checkpoint(ra_index(), term(), ra_server_state()) ->
+      {ra_server_state(), effects()}.
+
```
kjnilsson, line 1855: redundant blank line

src/ra_machine.erl
```diff
@@ -135,6 +135,8 @@
     {log, [ra_index()], fun(([user_command()]) -> effects())} |
     {log, [ra_index()], fun(([user_command()]) -> effects()), {local, node()}} |
     {release_cursor, ra_index(), state()} |
+    {release_cursor, ra_index()} |
```
kjnilsson, line 138: I think these two new effects need to be added to `ra_server:filter_follower_effects/1` else it may be that followers won't be able to emit these.

`filter_follower_effects/1` is an old function and arguably it would be better and more intuitive to do the filtering of which effects should and shouldn't be executed based on the Raft state in `ra_server_proc:handle_effects/`

src/ra_log.erl
```diff
@@ -668,23 +712,40 @@ update_release_cursor0(Idx, Cluster, MacVersion, MacState,
...
+                    undefined -> SnapInter;
+                    {I, _} -> I + SnapInter
+                end,
+    % The release cursor index is the last entry _not_ contributing
+    % to the current state. I.e. the last entry that can be discarded.
+    % We should take a snapshot if the new snapshot index would allow us
+    % to discard any segments or if the we've handled enough commands
+    % since the last snapshot.
+    CanFreeSegments = lists:any(fun({_, To, _}) -> To =< Idx end,
+                        ra_log_reader:segment_refs(Reader)),
```
kjnilsson, line 738: indentation?

src/ra_machine.erl
kjnilsson: Might be nice to get some kind of full integration test in for checkpoints, just to validate the happy path. Perhaps in `coordionation_SUITE`

src/ra_snapshot.erl
```diff
@@ -39,12 +39,22 @@
...
         ]).
 
 -type effect() :: {monitor, process, snapshot_writer, pid()}.
 
--export_type([meta/0, file_err/0, effect/0, chunk_flag/0]).
+-type kind() :: snapshot | checkpoint.
+
+-type checkpoint() :: ra_idxterm().
+
+-export_type([meta/0, file_err/0, effect/0, chunk_flag/0, kind/0,
```
kjnilsson, line 56: I would prefer exports to be formatted on their own lines

src/ra_log.erl
```diff
@@ -148,15 +156,20 @@ init(#{uid := UId,
     LogId = maps:get(log_id, Conf, UId),
     ResendWindow = maps:get(resend_window, Conf, ?DEFAULT_RESEND_WINDOW_SEC),
     SnapInterval = maps:get(snapshot_interval, Conf, ?SNAPSHOT_INTERVAL),
+    CPInterval = maps:get(checkpoint_interval, Conf, ?CHECKPOINT_INTERVAL),
```
kjnilsson, line 159: I wonder if we should take the opportunity to try to name these something like `min_checkpoint_interval` We'd have to keep `snapshot_interval` in ra_log_init_args() of but it could be in addition to `min_snapshot_interval`

kjnilsson (changes requested):
This is a very high quality PR - I did managed to find one thing will need addressing (which to be fair isn't particularly intuitive) as well as a couple of nitpicks.

src/ra_log.erl
the-mikedavis: Makes sense to me :+1:

I'll make a follow-up PR for adding `min_snapshot_interval` and falling back to `snapshot_interval`

kjnilsson (approved):

PR #425, opened 2024-03-08
Add reply_mode to command meta data

kjnilsson:
This would allow state machines to make use of, e.g. the pid in the `{notify, Corr, Pid}` reply mode to avoid having to include Pids in the command body and thus avoid them being stored twice for each command.


the-mikedavis (approved):
Great idea!

PR #426, opened 2024-03-12
Add a gauge for the effective machine version in ra_counters

### LINKED ISSUE #424: Add effective_machine_version to ra_counters as a guage

It could be handy to have this information easily queryable such that clients could use it to determine if a given command is supported or not.

the-mikedavis:
This allows callers to cheaply query a server's current effective machine version.

Closes #424
This should be useful for https://github.com/rabbitmq/khepri/pull/250

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -339,6 +339,7 @@ init(#{id := Id,
     put_counter(Cfg, ?C_RA_SVR_METRIC_COMMIT_INDEX, CommitIndex),
     put_counter(Cfg, ?C_RA_SVR_METRIC_LAST_APPLIED, SnapshotIdx),
     put_counter(Cfg, ?C_RA_SVR_METRIC_TERM, CurrentTerm),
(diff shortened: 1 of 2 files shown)


src/ra_server.erl
```diff
@@ -2557,6 +2558,7 @@ apply_with({Idx, Term, {noop, CmdMeta, NextMacVer}},
     IsOk = MacVer >= NextMacVer,
     case NextMacVer > OldMacVer of
         true when IsOk ->
+            put_counter(Cfg0, ?C_RA_SVR_METRIC_EFFECTIVE_MACHINE_VERSION, NextMacVer),
```
kjnilsson, line 2561: I think we may need to make sure this is set after snapshot installations also.

kjnilsson (changes requested):
One comment to look into. otherwise fine!

kjnilsson (approved):

Issue #416, opened 2024-02-05
`ra_log_cache_key_not_found` exception exit occured


sile:
### Describe the bug

The following exception was raised when processing a consistent query:

```
exception exit: {ra_log_cache_key_not_found,15}
      in function  ra_log_cache:fetch/2 (_build/default/lib/ra/src/ra_log_cache.erl, line 68)
      in call from ra_log:'-resend_from0/2-fun-0-'/3 (_build/default/lib/ra/src/ra_log.erl, line 932)
      in call from ra_log:'-resend_from0/2-lists^foldl/2-0-'/3 (_build/default/lib/ra/src/ra_log.erl, line 931)
      in call from ra_log:resend_from/2 (_build/default/lib/ra/src/ra_log.erl, line 915)
      in call from ra_log:handle_event/2 (_build/default/lib/ra/src/ra_log.erl, line 456)
      in call from ra_server:handle_follower/2 (_build/default/lib/ra/src/ra_server.erl, line 1123)
      in call from ra_server_proc:handle_follower/2 (_build/default/lib/ra/src/ra_server_proc.erl, line 1090)
      in call from ra_server_proc:follower/3 (_build/default/lib/ra/src/ra_server_proc.erl, line 794)
      in call from gen_statem:loop_state_callback/11 (gen_statem.erl, line 1395)
```

- ra version: 2.9.0

### Reproduction steps

I am unable to provide the reproduction steps as the chance of the exception occurring is very rare, and it happened while running test code for our non-open source product.
(Feel free to ignore this issue if you think there is insufficient information.)

Instead, here is a rough outline of the test scenario:
1. Construct a cluster consisting of 5 nodes, where each node is named as a, b, c, d, and e.
2. Regularly monitor the availability of the cluster by running a consistent query.
3. Divide the nodes into two groups: {a,b} forms one group, while {c,d,e} forms the other group.
   - We employ a custom [Erlang distribution module](erl_dist) that emulates a significantly slow network (when a certain flag is enabled, communication between the connected nodes is severely limited.)
4. Restore the split cluster to its normal state.

The exception mentioned above appears to have occurred at a node in the majority group when running a consistent query (for health check) just after the 3rd step.

[erl_dist]: https://www.erlang.org/doc/apps/erts/alt_dist

### Expected behavior

No exception exits happen.

kjnilsson:
we have seen this error a couple of times but not been able to trace it down. After the process has crashed and restarted it should be ok typically. Was this the case here?

[...]

PR #428, opened 2024-04-11
Improve Ra server resilience when log infrastructure encounters faults

kjnilsson:
Various improvements to data safety when log infrastructure processes encounter faults.

In particular there are many improvements and fixes relating to the server -> wal resend protocol including:
* Bug fix to ra_log_cache that would cause most triggered resends result in a ra process crash.
* Dropping fewer messages using the gen_state postpone feature.
* Ra leaders would previously just exit with `wal_down` - now they enter the same await_condition state although with a shorter timeout after which the begin a leader transfer process
* Improved detection and availability when a command is lost on the way to the wal and no further commands are sent.

Also there is a new feature to configure on a per system basis what kind of server recovery should take place when a ra system starts/restarts. There are 3 options:
1. `undefined` : do not restart any ra server
2. `registered`: restart all locally registered servers for the system
3. `mfa`: call a custom function that performs the restart.

This feature will allow dynamically started ra server to be restarted should the ra system crash and restart.

Also improvements to code coverage and refactoring.

Fixes: https://github.com/rabbitmq/ra/issues/416


pjk25:
@kjnilsson you can rebase now that #431 is merged

src/ra_server.erl
```diff
@@ -518,51 +521,76 @@ handle_leader({PeerId, #append_entries_reply{success = false,
...
+                           [{PeerId, _} | _] ->
+                               [{next_event, cast, {transfer_leadership, PeerId}}]
+                       end,
 
-    {leader, State, Effects};
+            State = State0#{condition =>
+                            #{predicate_fun => fun wal_down_condition/2,
+                              transition_to => leader,
+                              %% TODO: make duration configurable?
+                              timeout => #{duration => 5000,
```
michaelklishin, line 588: We have seen enough false positives with 5s timeouts that I'd suggest making it 10s even though any hardcoded default will be wrong/suboptimal for certain installations.

michaelklishin (approved):
I have pushed some mostly cosmetic changes:

 * Less logging
 * Clarified a few comments

src/ra_server.erl
michaelklishin: Actually, this should be enough time for `wal_down_condition/2` to observe a WAL process change, and even if it is observed at a later point, the resending can still do its job.

kjnilsson: yeah I think 5s is fine, it does affect - if we see ill effects we can make it configurable but I doubt it, worse case it  goes back to leader then have to go back to await_condition again.

PR #445, opened 2024-05-24
WAL: drop entries with indexes lower or equal to the last snapshot

kjnilsson:
In situations where the WAL lags behind but entries are still
committed the WAL will still write entries with indexes lower
than the last snapshot index for a given server. To avoid this
unnessary work we no check the ra_log_snapshot_state table for
each write and do not write them if they are lower than the
snapshot index. This allows the wal a better chance to catch
up on it's backlog.

Some changes needed to be made to the ra_log also to handle
this new behaviour.

### DIFF
--- a/test/ra_log_2_SUITE.erl
+++ b/test/ra_log_2_SUITE.erl
@@ -52,6 +52,7 @@ all_tests() ->
      sparse_read_out_of_range,
      sparse_read_out_of_range_2,
      written_event_after_snapshot,
+     writes_lower_than_snapshot_index_are_dropped,
      updated_segment_can_be_read,
      open_segments_limit,
      external_reader,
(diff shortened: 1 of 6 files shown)


test/ra_log_2_SUITE.erl
```diff
@@ -426,6 +427,62 @@ written_event_after_snapshot(Config) ->
...
+writes_lower_than_snapshot_index_are_dropped(Config) ->
+    Log0 = ra_log_init(Config, #{min_snapshot_interval => 1}),
+    Log1 = ra_log:append({1, 1, <<"one">>}, Log0),
+    Log1b = deliver_all_log_events(ra_log:append({2, 1, <<"two">>}, Log1), 500),
+    true = erlang:suspend_process(whereis(ra_log_wal)),
+    Log2 = write_n(3, 500, 1, Log1b),
+    {Log3, _} = ra_log:update_release_cursor(100, #{}, 1,
+                                             <<"100">>, Log2),
+    Log4 = deliver_all_log_events(Log3, 500),
+    %% TODO: assertion
```
the-mikedavis, line 439: Is this todo resolved with the `?assertMatch` below?

kjnilsson: yep!

PR #446, opened 2024-05-29
coordination_SUITE: Improve wait condition in `segment_writer_or_wal_crash_leader/1`

dumbbell:
## Why

The previous condition was not enough. This led to a failing assertion when the cache size was checked.

Submitted-by: @kjnilsson


dumbbell:
Karl provided the patch and granted me the permission to commit it and merge it.

PR #451, opened 2024-07-01
Emit the new `local_query` tuple only if query options are set

dumbbell:
## Why

Using the new tuple isn't compatible with older versions of Ra. This breaks local/leader queries if there are Ra members running an old version.

## How

We don't need the new `{local_query, QueryFun, Options}` tuple if there is no options. We can use the old `{local_query, QueryFun}` in this case and keep the compatibility.


dumbbell:
That's correct, I don't think it requires a minor bump.

PR #455, opened 2024-07-12
Do not update ra_log_snapshot_state from snapshot writer process.

kjnilsson:
As this could cause entries not to be written to the WAL (as it thinks they will never be read) but the main ra process is not yet aware of the snapshot and thus may try to read these entries.

This line was never necessary, it was an attempt at an optimisation for the checkpoints feature.

also:

Clear out older checkpoint when a snapshot is installed else this could cause old checkpoints to be promoted.


mkuratczyk (approved):
We've tested the same workload that had triggered the issue before and we can no longer reproduce it, even after "helping" manually (`ra:local_query(FollowerId, fun (_) -> timer:sleep(10000) end)`).

PR #456, opened 2024-07-18
Fixes for pre-init recovery and handling of gaps in mem tables due to snapshotting

kjnilsson:
The first thing a ra system does when it starts is run a pre-init
phase for each registered Ra server, mostly to recover the
ra_log_snapshot_state table. This appears to have been broken
for along time and the ra_log_snapshot_table has not been populated
ahead of WAL / segment writer recovery. This was fine as this bit
was just an optimisation and never affected the workings of
the Ra infrastructure.

However since https://github.com/rabbitmq/ra/commit/5b7a26555b6aeefda143297d92e543e276046d07 it needs
this in order to avoid the segment writer crashing when it detects
a gap (caused by the WAL dropping entries lower than the current
snapshot).

This commit mostly fixes the pre-init process but also addresses
a potential race condition which still could cause the segment
writer to crash for the same reason.

See: https://github.com/rabbitmq/rabbitmq-server/discussions/11712

### DIFF
--- a/src/ra_log_segment_writer.erl
+++ b/src/ra_log_segment_writer.erl
@@ -308,40 +308,57 @@ append_to_segment(_, _, StartIdx, EndIdx, Seg, Closed, _State)
   when StartIdx >= EndIdx ->
     {Seg, Closed};
 append_to_segment(UId, Tid, Idx, EndIdx, Seg0, Closed, State) ->
-    [{_, Term, Data0}] = ets:lookup(Tid, Idx),
-    Data = term_to_iovec(Data0),
-    DataSize = iolist_size(Data),
-    case ra_log_segment:append(Seg0, Idx, Term, {DataSize, Data}) of
-        {ok, Seg} ->
-            ok = counters:add(State#state.counter, ?C_ENTRIES, 1),
-            %% this isn't completely accurate as firstly the segment may not
-            %% have written it to disk and it doesn't include data written to
-            %% the segment index but is probably good enough to get comparative
-            %% data rates for different Ra components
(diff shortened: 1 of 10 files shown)


src/ra_log_segment_writer.erl
```diff
@@ -308,40 +308,57 @@ append_to_segment(_, _, StartIdx, EndIdx, Seg, Closed, _State)
...
+                  [Idx, Tid, UId]),
+            case ra_directory:is_registered_uid(State#state.system, UId) of
+                true ->
+                    ?ERROR("segment_writer: uid ~s is registered, exiting...",
+                           [UId]),
+                    exit({missing_index, UId, Idx});
+                false ->
+                    ?INFO("segment_writer: UId ~s was not registered, skipping",
+                          [UId]),
+                    throw(undefined)
```
the-mikedavis, line 325: Is the `catch/1` above wrapping the `append_to_segment/6` call only meant to catch this throw? I think the `throw(undefined)` can just be a return of `undefined` here and we could remove the `catch/1` above. The `catch/1` will also catch the exits in the other branches which will be a change of behavior. It looks like that might lead to an unexpected match since `{'EXIT', Reason}` will be matched as `{Segment, Closed0}` above and then a bad generator error

the-mikedavis:
tiny typo in the commit message & description: apperas => appears

src/ra_log_segment_writer.erl
kjnilsson: not sure what I was thinking with that throw :) it was late when I wrote it

kjnilsson:
> tiny typo in the commit message & description: apperas => appears

fixed

PR #464, opened 2024-08-08
Kill snapshot process before deleting everything

kjnilsson:
if there is a snapshot process pending it could cause the directory deletion to fail, best kill the snapshot process first

### DIFF
--- a/src/ra_log.erl
+++ b/src/ra_log.erl
@@ -933,8 +933,23 @@ read_config(Dir) ->
     ra_lib:consult(ConfigPath).
 
 -spec delete_everything(state()) -> ok.
-delete_everything(#?MODULE{cfg = #cfg{directory = Dir}} = Log) ->
+delete_everything(#?MODULE{cfg = #cfg{directory = Dir},
+                           snapshot_state = SnapState} = Log) ->
     _ = close(Log),
+    %% if there is a snapshot process pending it could cause the directory
+    %% deletion to fail, best kill the snapshot process first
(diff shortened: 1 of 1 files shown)


src/ra_log.erl
```diff
@@ -933,8 +933,24 @@ read_config(Dir) ->
...
+                           snapshot_state = SnapState} = Log) ->
     _ = close(Log),
+    %% if there is a snapshot process pending it could cause the directory
+    %% deletion to fail, best kill the snapshot process first
+    case ra_snapshot:pending(SnapState) of
+        {Pid, _, _} ->
+            case is_process_alive(Pid) of
+                true ->
+                    exit(Pid, kill),
+                    timer:sleep(100),
```
the-mikedavis, line 946: I doubt the timing would be an issue but I wonder if we should avoid the sleep here and instead monitor the process and await the `DOWN` notification instead? Both `erlang:monitor/2` and `erlang:exit/2` should handle the process already having exited gracefully:

```erl
case ra_snapshot:pending(SnapState) of
    {Pid, _, _} ->
        MRef = erlang:monitor(process, Pid),
        true = erlang:exit(Pid, kill),
        receive
            {'DOWN', MRef, process, Pid, _Reason} ->
                ok
        after
            500 ->
                %% log an error?
        end;
    _ ->
        ok
end
```

kjnilsson: yeah I dont think we need to sleep or check anything.

PR #468, opened 2024-09-04
Amend ra_machine:snapshot_installed/4

kjnilsson:
Instead of passing the plain old version we build a full snapshot meta so that we can also compare index, term and membership as well as the machine version.

Changed the order of arguments.

Removed the optional `snapshot_installed/2` callback but retained
the fallback call in ra_machine in case it is still implemented
somwhere.

### DIFF
--- a/src/ra_machine.erl
+++ b/src/ra_machine.erl
@@ -209,7 +209,6 @@
               command_meta_data/0]).
 
 -optional_callbacks([tick/2,
-                     snapshot_installed/2,
                      snapshot_installed/4,
                      state_enter/2,
                      init_aux/1,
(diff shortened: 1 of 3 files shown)


dumbbell (approved):
Looks good to me. @the-mikedavis, could you please take a look as well?

the-mikedavis (approved):

src/ra_machine.erl
```diff
@@ -242,8 +241,6 @@
 
 -callback tick(TimeMs :: milliseconds(), state()) -> effects().
 
--callback snapshot_installed(ra_snapshot:meta(), state()) -> effects().
-
 -callback snapshot_installed(Meta, OldMacVer, OldState, NewState) -> Effects
```
the-mikedavis, line 244: Looks like this needs an update to be `State, Meta, OldState, OldMeta`

the-mikedavis (changes requested):

src/ra_machine.erl
kjnilsson: oops yea

the-mikedavis (approved):

PR #470, opened 2024-09-11
ra_server: handle higher-term AERs in `receive_snapshot` state

keynslug:
## Proposed Changes

When in `receive_snapshot` state, server should handle at least higher-term AERs gracefully. Otherwise, server could become stuck in the `receive_snapshot` state if the current snapshot sender goes away and the new leader gets elected, before snapshot transfer is complete. In this case the new leader will keep sending AERs to this stuck server, which it will ignore but (unfortunately) keep resetting `receive_snapshot_timeout` each time. This constant timer reset prevents `receive_snapshot_timeout` timer to fire, unless `receive_snapshot` timeout option is impractically small.

## Types of Changes

- [x] Bug fix (non-breaking change, no corresponding GH issue)

## Checklist

- [ ] ~~Any dependent changes have been merged and published in related repositories~~

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -1411,6 +1411,16 @@ handle_receive_snapshot(#install_snapshot_rpc{term = Term,
             State = update_term(Term, State0#{log => Log}),
             {receive_snapshot, State, [{reply, Reply}]}
     end;
+handle_receive_snapshot(#append_entries_rpc{term = Term} = Msg,
+                        #{current_term := CurTerm,
+                          cfg := #cfg{log_id = LogId}} = State)
+  when Term > CurTerm ->
+    ?INFO("~ts: follower receiving snapshot saw append_entries_rpc from ~w for term ~b "
+          "abdicates term: ~b!",
+          [LogId, Msg#append_entries_rpc.leader_id,
+           Term, CurTerm]),
+    {follower, update_term(Term, clear_leader_id(State)),
+     [{next_event, Msg}]};
 handle_receive_snapshot({ra_log_event, Evt},
                         State = #{cfg := #cfg{id = _Id, log_id = LogId},
                                   log := Log0}) ->
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -1413,13 +1413,17 @@ handle_receive_snapshot(#install_snapshot_rpc{term = Term,
     end;
 handle_receive_snapshot(#append_entries_rpc{term = Term} = Msg,
                         #{current_term := CurTerm,
-                          cfg := #cfg{log_id = LogId}} = State)
+                          cfg := #cfg{log_id = LogId},
+                          log := Log0} = State)
   when Term > CurTerm ->
     ?INFO("~ts: follower receiving snapshot saw append_entries_rpc from ~w for term ~b "
           "abdicates term: ~b!",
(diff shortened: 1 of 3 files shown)


kjnilsson:
Thanks for this, it looks ok to me. I will take a closer look tomorrow.

the-mikedavis (commented):
Could you add a test case for this branch? https://github.com/rabbitmq/ra/blob/14eca5fe898e9ad7c39abe82d68cff82b7002c9d/test/ra_server_SUITE.erl#L2086-L2108 can be mostly reused and we can swap out the `receive_snapshot_timeout` message for an AER from the next term

I've pushed some changes to the CI so if you rebase on `main` I believe the CI should pass on the next run

src/ra_server.erl
```diff
@@ -1411,6 +1411,16 @@ handle_receive_snapshot(#install_snapshot_rpc{term = Term,
...
     end;
+handle_receive_snapshot(#append_entries_rpc{term = Term} = Msg,
+                        #{current_term := CurTerm,
+                          cfg := #cfg{log_id = LogId}} = State)
+  when Term > CurTerm ->
+    ?INFO("~ts: follower receiving snapshot saw append_entries_rpc from ~w for term ~b "
+          "abdicates term: ~b!",
+          [LogId, Msg#append_entries_rpc.leader_id,
+           Term, CurTerm]),
+    {follower, update_term(Term, clear_leader_id(State)),
```
kjnilsson, line 1422: I think we need to call `ra_snapshot:abort_accept/1` here?

kjnilsson:
I think we should add some more changes on how the `state_timeout` is reset to avoid to being reset for any message that isn't snapshot related. So that [this action](https://github.com/rabbitmq/ra/blob/0146f87f599e4c8f9ddc10a209cb46137f54cd60/src/ra_server_proc.erl#L877-L880) is only emitted when an install_snapshot_rpc message is received. I think it would be fine to check the message type in `ra_server_proc` at least for now.

src/ra_server_proc.erl
```diff
@@ -874,10 +874,16 @@ receive_snapshot(EventType, Msg, State0) ->
         {receive_snapshot, State1, Effects} ->
             {#state{conf = Conf} = State, Actions} =
                 ?HANDLE_EFFECTS(Effects, EventType, State1),
-            #conf{receive_snapshot_timeout = ReceiveSnapshotTimeout} = Conf,
-            {keep_state, State,
-             [{state_timeout, ReceiveSnapshotTimeout,
-               receive_snapshot_timeout} | Actions]};
+            TimeoutActions = case Msg of
```
kjnilsson, line 877: This is fine but can you format it like the rest of the code in the module. `case` and `end` should line up.

kjnilsson (approved):
Just a formatting change.

src/ra_server_proc.erl
keynslug: Fixed up in 7851f9182ade17f9fd25da1b6e7ca52a6536f502.

michaelklishin:
The RabbitMQ OCI build failure is due to the fact that external contributions do not have access to GitHub secrets. It has nothing to do with the code changes.

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #487, opened 2024-12-11
Optimise read planning

### LINKED ISSUE #486: Optimise segment sparse reads for external readers.

Currently when opening a segment for reading the entire index is rebuilt and kept in memory. For the default segment size of 4096 entries this index takes up ~42Kb of memory. Also external reads ought to always rebuild the index for segments that aren't full every time they are requested to read to avoid accidentally reading an overwritten entry in a prior term (all entries are appended to segments, no actual overwrites occur and resolving the correct entry to read is done at index parse time).

Hence it would make sense to profile/measure and optimise segment open and index parsing and index size.

A couple of options:

Instead of a map of `#{ra_index() => {ra_term(), Offset, Length, Crc}}` we could compact the index for external readers to something like:

```
[[{ra_index(), Offset, Length}...]]

```

where the inner rows are of max fixed length, e.g 64 entries. this would then require a maxiumum of 64+64 operations for a read which is better than keeping a flat list but more space efficient (~24Kb) than a map.

Another option:

Parse the binary index on demand. This could result in too many file syscalls but would be a good option for segments that arent full perhaps as these need to be re parsed on every read request.

### Describe alternatives you've considered

.

### Additional context

.

kjnilsson:
Previously we had to do do a linear search through the list of
segment reference to find which file to read a given entry from.

This commit introduced a new module `ra_lol` which is a
"list of list" type of ordered data structure that replaces
the use of a plain list.

ra_lol keeps data in a list of 64 entry sized "rows" which
subtantially reduces the amount of work needed to find a
given segment ref as we first scan each row to find the
row with the segref in then we can find the fixed size row.

Even a server with 4096 segment only need to scan at most
128 to find any given entry.

This commit also fixes a few issues in segref compaction
and changes the type of the segment ref to a {ra_range:range(), filename()}
to work better with the ra_range module.

Fixes #486

### DIFF
--- a/src/ra_log_reader.erl
+++ b/src/ra_log_reader.erl
@@ -15,6 +15,8 @@
          update_segments/2,
          handle_log_update/2,
          segment_refs/1,
+         segment_ref_count/1,
+         range/1,
          num_open_segments/1,
          update_first_index/2,
          fold/5,
@@ -36,12 +38,10 @@
               access_pattern = random :: access_pattern()
              }).
 
--type segment_ref() :: {From :: ra_index(), To :: ra_index(),
-                        File :: string()}.
+-type segment_ref() :: ra_log:segment_ref().
 -record(?STATE, {cfg :: #cfg{},
-                 % first_index = 0 :: ra_index(),
-                 last_index = 0 :: ra:index(),
-                 segment_refs = [] :: [segment_ref()],
+                 range :: ra_range:range(),
+                 segment_refs :: ra_lol:state(),
                  open_segments :: ra_flru:state()
                 }).
 
@@ -52,7 +52,7 @@
 -export_type([
               state/0,
               read_plan/0
-              ]).
+             ]).
 
 %% PUBLIC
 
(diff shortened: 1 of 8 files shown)


src/ra_log_reader.erl
```diff
@@ -95,18 +101,30 @@ close(#?STATE{open_segments = Open}) ->
...
+    SegmentRefsComp = compact_segrefs(NewSegmentRefs, SegmentRefs0),
+    SegmentRefsCompRev = lists:reverse(SegmentRefsComp ),
+    SegRefs = ra_lol:from_list(fun seg_ref_gt/2, SegmentRefsCompRev),
+    Range = case SegmentRefsComp of
+                  [{{_, L}, _} | _] ->
+                      [{{F, _}, _} | _] = SegmentRefsCompRev,
+                      ra_range:new(F, L);
+                  _ ->
+                      undefined
+              end,
```
the-mikedavis, line 116: minor formatting thing: I think this block needs one or two columns of indent removed

PR #491, opened 2024-12-20
Refactor consistent query

### LINKED ISSUE #483: consistent query bug
### Describe the bug

https://github.com/rabbitmq/ra-kv-store/actions/runs/12121409210

In. one test (in a long history of non failing runs) one leader in a minority partition decided it was ok to serve consistent reads.
After a fair bit of time investigating it wasn't clear exactly what caused it but I suspect it was due to the leader somehow ending up with an incorrectly high query_index for one of it's peers.

The consistent query logic can be refactored to avoid relying on this piece of state entirely as all it does is perform a leader liveness check then waits for the commit index to be appled then execute the query.

This check can be done without each member keeping a query_index and instead just returning the {term, index} tuple provided _if_ the term is equal or higher.

### Reproduction steps

https://github.com/rabbitmq/ra-kv-store/actions/runs/12121409210

### Expected behavior

.

### Additional context

```erlang
handle_follower(#heartbeat_rpc{query_index = RpcQueryIndex, term = Term,
                               leader_id = LeaderId},
                #{current_term := CurTerm,
                  cfg := #cfg{id = Id}} = State0)
  when Term >= CurTerm ->
    State1 = update_term(Term, State0),
    #{query_index := QueryIndex} = State1,
    NewQueryIndex = max(RpcQueryIndex, QueryIndex),
    State2 = update_query_index(State1#{leader_id => LeaderId}, NewQueryIndex),
    Reply = heartbeat_reply(State2),
    {follower, State2, [cast_reply(Id, LeaderId, Reply)]};
```

In this code, if the follower somehow ended up with a local_query index that was higher than what the leader has (each term resets the query index) then that could cause a leader to locally store a higher query index which if this follower was later partition would still contribute to the quorum calculation. If we can remove this state we can remove this possibility.

kjnilsson:
Instead of each member maintaining a shared query_index updated by the heartbeat_rpc messages this commit changes the query_index to just maintain each member's local sequence that is only used when the member is a leader.

This reduces some complexity in the algorithm to possibly avoids certain bugs that could result from a follower retaining an old higher query index.

May fix #483

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -731,11 +731,13 @@ handle_leader(#heartbeat_rpc{term = Term} = Msg,
            Term, CurTerm]),
     {follower, update_term(Term, State0#{leader_id => undefined}),
      [{next_event, Msg}]};
-handle_leader(#heartbeat_rpc{term = Term, leader_id = LeaderId},
+handle_leader(#heartbeat_rpc{term = Term,
+                             query_index = QueryIndex,
+                             leader_id = LeaderId},
               #{current_term := CurTerm,
                 cfg := #cfg{id = Id}} = State)
         when CurTerm > Term ->
-    Reply = heartbeat_reply(State),
+    Reply = heartbeat_reply(CurTerm, QueryIndex),
     {leader, State, [cast_reply(Id, LeaderId, Reply)]};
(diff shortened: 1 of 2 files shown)


kjnilsson:
Passes two runs of the jepsen test suite.

PR #493, opened 2024-12-24
Delay machine upgrade until all members are up-to-date

### LINKED ISSUE #490: Delay machine version upgrade until all members have the new version

Currently the machine version is incremented when a leader with the new version is elected and this upgrade is communicated via the noop command that the leader persists. A leader with a higher version can only be elected once it has a quorum but this can leave a minority of member unable to apply command (although they participate in replication / quorum). This can cause problems with local data freshness and for RabbitMQ quorum queue messsage deliveries on minority nodes will stop until the node is upgraded.

Delay upgrading the machine version until the leader knows all members have the right machine version available.

The upgrade still needs to be done when handing a command communicating the upgrade but we could delay committing this command until the pre-requisite is met

1. Add a new protocol record pair: `info_rpc{}` and `info_reply{}` - these "RPC"s are async like most other protocol commands. `info_rpc` will take a list of info keys that the leader wants other nodes to provide with initally only a single key is supported `machine_version`. `info_reply` will include a map of the requestioned information.
2. The leader will emit info_rpc requests to all followers when it becomes a leader (possibly only when it becomes a leader and a higher machine version is available). The leader will gather the replies in order to discover the machine version of each member and store this in the corresponding peer maps.
The leader will also emit an info_rpc every tick for each member for which the machine version is unknown or that has a machine version that is lower that the leader's.
3. If the "delay machine version upgrade" feature is enabled the leader will not commit its current machine version in the [noop command](https://github.com/rabbitmq/ra/blob/ece627aa1dc5fae217f26a5f0a601501f181bf91/src/ra_server.erl#L872-L873) (which is the first command in every new term). Instead it will just use the current effective machine version.
4. Once the leader has established that all members have the same new machine version it commits another noop command with the current version which will trigger the upgrade.

dumbbell:
## Why

Before this patch, a Ra cluster would switch to a new machine version immediately after a leader with that version was elected.

Because a leader can be elected with a quorum number of candidate voting for it, it means the cluster could start using the new machine version as soon as a quorum of members support that version.

Unfortunately, other members that do not support it stop applying commands because they run an older version of the machine code. For some consumers of Ra, like Khepri, this means they could cease their operation locally until the member is restarted with the new machine version.

We want to delay the machine upgrade to a point where all members know about the new version. This ensures all members can continue to provide their service.

## How

The machine version to use is communicated by the leader using the `noop` command. This command is the first one sent just after an election. The machine version passed was the local machine version.

With this patch, the `noop` command sent after an election passes the effective machine version, except if the leader is unclustered alone (in which case it passes the latest machine version. Therefore in a cluster, the leader will send a second `noop` command with a newer machine version later, once all members support it.

To determine what each follower supports, this patch introduces two commands:
* `#info_rpc{}`
* `#info_reply{}`

Once a leader is elected, in addition to the `noop` command, it sends an `#info_rpc{}` command to all followers. They reply with `#info_reply{}` with the machine version they support. This mechanism is not specific to machine upgrades: this could be extended in the future to communicate more details about each follower.

Once the leader received the machine version of every followers, it can determine the highest possible supported machine version. For that, it simply takes the lowest reported machine version (including the leader's machine version). If this version is greater than the effective machine version, the leader sends a new `noop` command with the new machine version to use.

The leader sends the `#info_rpc{}` command again and again to some followers at each "tick", if these followers did not report anything yet, or if the reported machine version is lower than its own supported machine version. This takes care of follower that did not receive the initial `#info_rpc{}` and those that were restarted as part of an upgrade.

Fixes #490.


src/ra.hrl
```diff
@@ -187,6 +188,17 @@
...
+-record(info_rpc,
+        {from :: ra_server_id(),
+         term :: ra_term(),
+         keys :: [ra_server:ra_server_info_key()]}).
+
+-record(info_reply,
+        {from :: ra_server_id(),
+         term :: ra_term(),
+         keys :: [ra_server:ra_server_info_key()],
+         info :: ra_server:ra_server_info() | undefined}).
```
kjnilsson, line 200: default this to `#{}` and remove the `undefined` option?

src/ra_server.erl
```diff
@@ -3314,6 +3443,186 @@ after_log_append_reply(Cmd, Idx, Term, Effects0) ->
...
+            Id = id(State),
+            Command = #info_rpc{from = Id,
+                                term = CurTerm,
+                                keys = [machine_version]},
+            [{send_rpc, PeerId, Command}];
+        false ->
+            []
+    end.
+
+info_reply_effect(
```
kjnilsson, line 3517: this should fit to the right of `(` so no need for newline here.

src/ra_server.erl
```diff
@@ -3314,6 +3443,186 @@ after_log_append_reply(Cmd, Idx, Term, Effects0) ->
...
+        [] ->
+            %% This node is alone in the cluster, we can send the `noop'
+            %% command with the newer machine version right away.
+            MacVer = ra_machine:version(Mac),
+            Noop = {noop, #{ts => erlang:system_time(millisecond)},
+                    MacVer},
+            NoopEffect = {next_event, cast, {command, Noop}},
+            [NoopEffect];
+        _ ->
+            %% We continue to send the `noop' command immediately after
```
kjnilsson, line 3461: I am wondering if we should build in a mechanism to keep the old behaviour

src/ra_server.erl
```diff
@@ -3314,6 +3443,186 @@ after_log_append_reply(Cmd, Idx, Term, Effects0) ->
...
+              end, Info0),
+    Info1.
+
+handle_info_reply(
+  #{cluster := Cluster} = State,
+  #info_reply{from = PeerId, keys = Keys, info = Info}) ->
+    PeerState0 = maps:get(PeerId, Cluster),
+    PeerState1 = case Info of
+                     undefined ->
+                         %% The peer is in a Raft state where it can't give
```
kjnilsson, line 3557: I think a peer should always be able to provide info for keys it understands.

src/ra_server.erl
```diff
@@ -3314,6 +3443,186 @@ after_log_append_reply(Cmd, Idx, Term, Effects0) ->
...
+                           end, PeerState0, Keys)
+                 end,
+    Cluster1 = Cluster#{PeerId => PeerState1},
+    State1 = State#{cluster => Cluster1},
+    determine_if_machine_upgrade_allowed(State1).
+
+determine_if_machine_upgrade_allowed(
+  #{cfg := #cfg{effective_machine_version = EffectiveMacVer,
+                log_id = LogId}} = State) ->
+    Effects = case has_enough_peer_info(State) of
```
kjnilsson, line 3577: do we need `has_enough_peer_info` at all if the later `get_max_supported_machine_version` could handle the case where some peers do not yet have a common machine version?

kjnilsson (changes requested):
looks good, a few tweaks and a question as to whether we should optionally keep the old behaviour or not.

src/ra_server.erl
dumbbell: Like an option in the Ra server config to perform the machine upgrade as soon as a leader with the new version is elected as before?

src/ra_server.erl
dumbbell: Ok, so I should get rid of the `empty_info_reply_effect()` function and let a Ra server always send its machine version in the `#info_reply{}`, correct?

src/ra_server.erl
kjnilsson: yes - just in case someone isn't ready to change to the new behaviour yet or we want to switch back for QQ for example. I'm thinking we could make it a ra system config key rather than a server config key.

src/ra_server.erl
kjnilsson: yes I think that makes sense,

dumbbell: In fact, disregard that previous comment.

src/ra.hrl
dumbbell: Done

src/ra_server.erl
dumbbell: Done

src/ra_server.erl
dumbbell: Addressed too :)

src/ra_server.erl
dumbbell: I merged the two functions.

src/ra_server.erl
dumbbell: Thank, let me change that! Any preferences for the name? I’m not satisfied with the `machine_upgrade_when` I came up with.

dumbbell: So I moved the option from the server to the system. The name is still `upgrade_machine_when` with two possible values: `after_election` or `when_available_everywhere`, the latter being the default.

That default is set in `src/ra_server.hrl`. `ra_system` queries the application environment and falls back to the default as part of the default system configuration.

src/ra_server.erl
```diff
@@ -192,6 +194,18 @@
 %% The simple machine config is version that can only be used for simple state
 %% machines that cannot access any of the advanced features.
 
+-type upgrade_machine_when() :: after_election | when_available_everywhere.
```
kjnilsson, line 197: we could consider `-type machine_upgrade_strategy() :: all | quorum.`

dumbbell: Suggestion applied :)

kjnilsson (approved):

PR #494, opened 2025-01-02
Ra v3

kjnilsson:
PR for visibility of v3 work.

See: https://github.com/rabbitmq/ra/milestone/7

### DIFF
--- a/src/ra.erl
+++ b/src/ra.erl
@@ -44,16 +44,10 @@
          delete_cluster/1,
          delete_cluster/2,
          % server management
-         % deprecated
-         start_server/1,
          start_server/2,
          start_server/5,
-         % deprecated
-         restart_server/1,
          restart_server/2,
          restart_server/3,
-         % deprecated
-         stop_server/1,
          stop_server/2,
          force_delete_server/2,
          trigger_election/1,
@@ -68,8 +62,6 @@
          leave_and_delete_server/3,
          leave_and_delete_server/4,
          %% troubleshooting
-         % deprecated
-         overview/0,
          overview/1,
          %% helpers
          new_uid/1,
@@ -78,19 +70,11 @@
          %% auxiliary commands
          aux_command/2,
          cast_aux_command/2,
-         register_external_log_reader/1,
          member_overview/1,
          member_overview/2,
          key_metrics/1,
          key_metrics/2
         ]).
-
-%% xref should pick these up
--deprecated({start_server, 1}).
--deprecated({restart_server, 1}).
--deprecated({stop_server, 1}).
--deprecated({overview, 0}).
--deprecated({register_external_log_reader, 1}).
 
 -define(START_TIMEOUT, ?DEFAULT_TIMEOUT).
 
@@ -170,19 +154,6 @@ start(Params) when is_list(Params) ->
 start_in(DataDir) ->
     start([{data_dir, DataDir}]).
 
-%% @doc Restarts a previously successfully started ra server in the default system
-%% @param ServerId the ra_server_id() of the server
-%% @returns `{ok | error, Error}' where error can be
-%% `not_found', `system_not_started' or `name_not_registered' when the
-%% ra server has never before been started on the Erlang node.
-%% DEPRECATED: use restart_server/2
-%% @end
--spec restart_server(ra_server_id()) ->
-    ok | {error, term()}.
-restart_server(ServerId) ->
-    %% TODO: this is a bad overload
-    restart_server(default, ServerId).
-
 %% @doc Restarts a previously successfully started ra server
 %% @param System the system identifier
 %% @param ServerId the ra_server_id() of the server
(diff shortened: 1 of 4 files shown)


src/ra_log_segments.erl
```diff
@@ -0,0 +1,1038 @@
...
+
+         %% perform sym linking of the additional segments in the compaction
+         %% group, the target is not yet updated which can be detected at
+         %% recovery by the presence of a sym link _and_ the .compacting
+         %% file
+         ok = make_symlinks(Dir, FirstSegmentFn,
+                            [F || {_, _, {F, _}} <- Additional]),
+
+         %% rename the .compacting segment on top of the group leader
+         ok = prim_file:rename(CompactingFn, FirstSegmentFn),
```
michaelklishin: Should we add an `fsync(2)` for the directory? Renaming a file on `ext4` is not guaranteed to be atomic without it *in some configurations* (but modern defaults seem to make this a nice-to-have).

Not sure how expensive it can be on a directory with a large number of segment files but this is a background operation, so should be acceptable.

docs/internals/COMPACTION.md
```diff
@@ -0,0 +1,320 @@
...
+
+This document describes the log compaction feature in Ra v3.
+
+## Overview
+
+Compaction in Ra is intrinsically linked to the snapshotting feature. Standard 
+Raft snapshotting removes all entries in the Ra log below the snapshot index 
+where the snapshot is a full representation of the state machine state.
+
+The high level idea of compacting in Ra is that instead of deleting all segment
```
michaelklishin: This does not really explain the key benefit of compaction: reclaiming disk space much more efficiently even if one consumer keeps holding on to a message with a very old Raft log entry, something that currently plagues QQs enough to [warrant a protection mechanism](https://www.rabbitmq.com/docs/consumers#acknowledgement-timeout) outside of Ra/QQs.

src/ra_log_segments.erl
kjnilsson: done

docs/internals/COMPACTION.md
kjnilsson: Ra doesn't have a concept of consumers but yes this doc could be read better for sure.

kjnilsson: updated doc with ra_kv example to better illustrate compaction use case.

src/ra_snapshot.erl
```diff
@@ -656,6 +761,27 @@ take_extra_checkpoints(#?MODULE{checkpoints = Checkpoints0,
             {State0, Checks}
     end.
 
+-spec write_indexes(file:filename_all(), ra_seq:state()) ->
+    ok | {error, file:posix()}.
+write_indexes(Dir, Indexes) ->
+    File = filename:join(Dir, <<"indexes">>),
+    ra_lib:write_file(File, term_to_binary(Indexes)).
```
ansd: I wonder whether the indexes file should also include a CRC32 checksum. Maybe not given, that those live indexes can be computed from the other snapshot file? On the other hand, the indexes file seems to get used upon recovery.

kjnilsson: yes good point, I will do that

kjnilsson: done

docs/internals/COMPACTION.md
```diff
@@ -0,0 +1,322 @@
...
+#{major_strategy => {num_minors, 5}}  %% trigger major after 5 minor compactions
+#{major_strategy => manual}           %% only manual triggering via ra:trigger_compaction/1
+```
+
+![segments](compaction2.jpg)
+
+### Phase 4 compaction (optional, not implemented)
+
+At some point the number of live indexes could become completely sparse (no
+adjacent indexes) and large which is sub-optimal memory wise.
```
the-mikedavis: Should we add a gauge to track sparse segments? I'm thinking about how we would know whether or not phase 4 compaction would be worth implementing - how we can detect how common this scenario would be

src/ra_flru.erl
```diff
@@ -91,6 +92,10 @@ evict_all(#?MODULE{items = Items,
 size(#?MODULE{items = Items}) ->
     length(Items).
 
+-spec max_size(state()) -> non_neg_integer().
+max_size(#?MODULE{max_size = MaxSize}) ->
+ MaxSize.
```
the-mikedavis:
```suggestion
    MaxSize.
```

src/ra_kv.erl
```diff
@@ -0,0 +1,297 @@
...
+    case maps:is_key(Key, Keys) of
+        true ->
+            State = State0#?STATE{keys = maps:remove(Key, Keys)},
+            {State, {ok, Meta}, []};
+        false ->
+            {State0, {error, not_found}, []}
+    end.
+
+live_indexes(#?STATE{keys = Keys}) ->
+    maps:fold(fun (_K, [Idx | _], Acc) ->
```
the-mikedavis:
```suggestion
    maps:fold(fun (_K, ?TUPLE(Idx, _), Acc) ->
```

for consistency with the other uses of the value part of this map

src/ra_lol.erl
the-mikedavis: There are a few files like this that are missing the license notice at the top

docs/internals/COMPACTION.md
```diff
@@ -0,0 +1,322 @@
...
+   using `ra_seq:in_range/2`
+3. Segments with zero live indexes are marked as "unreferenced" and deleted
+
+Minor compaction runs synchronously in the Ra server process as it only 
+involves checking ranges and deleting files - no data copying.
+
+**Special case**: If `LiveIndexes == []` (standard Raft behaviour), ALL 
+compactable segments are deleted immediately without checking ranges.
+
+### Major compaction (Phase 3)
```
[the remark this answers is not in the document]
kjnilsson: well.. it isn't missing per se, just not implemented yet. Added a section to that effect.

Issue #495, opened 2025-01-02
Avoid segment file name wrap around.


kjnilsson:
### Describe the bug

segment filenames are limited to 8 characters and are currently a base 10 number that is incremented every time a new segment is created. Although sufficient for most use cases it does mean that for very high throughput Ra clusters it is possible that the number will wrap around after a few years when creating a new segment every second or so.

A wrap around event would cause the segments to be sorted in the wrong order and would cause incorrect log reads.

### Expected behavior

by adding a prefix to the segment name we can ensure it is much harder to cause a wrap around event.

Alternative 1: Add the expansion of "~YYMMDD" to the segment name as a prefix. This would allow the creation of over 1000 segment files per second which is extremely unlikely. This approach would also retain the sorting requirements even for servers with pre-existing segment files with the old format.

This alternative creates quite odd looking file names

Alternative 2:

upgrade to a 16 or 20 character name and rename existing segments. this can be done in the segment writer's init function and a marker file can be written to avoid re-entering the upgrade flow once done.

This also should be done for the WAL.

PR #496, opened 2025-01-03
Extend WAL and segment file names from 8 to 16 characters

kjnilsson:
In order to avoid running out of characters after a few years of consistent usage.

The renames happens when the WAL and segment writer initialises respectively. The segment writer writes a marker after a successful upgrade in order to avoid listing files in all server directories every time. The WAL always checks the current set of wal files which should be ok given there is rarely more than one.

Fixes: #495


src/ra_log_segment_writer.erl
```diff
@@ -529,3 +532,26 @@ maybe_wait_for_segment_writer(SegWriter, TimeRemaining)
...
+            [begin
+                 Dir = filename:join(DataDir, UId),
+                 case file:list_dir(Dir) of
+                     {ok, Files} ->
+                         [ra_lib:zpad_upgrade(Dir, F, ".segment")
+                          || F <- Files, filename:extension(F) =:= ".segment"];
+                     {error, enoent} ->
+                         ok
+                 end
+             end || {_, UId} <- ra_directory:list_registered(System)],
```
the-mikedavis, line 551: Question to check my understanding: one `ra_log_segment_writer` is spawned per system, right? So it should be safe to upgrade everything in the segment writer's system?

test/ra_log_segment_writer_SUITE.erl
```diff
@@ -348,10 +348,11 @@ accept_mem_tables_for_down_server(Config) ->
     end,
     %% validate fake uid entries were written
     ra_log_segment_writer:await(?SEGWR),
-    FakeSegmentFile = filename:join([?config(wal_dir, Config),
-                                     FakeUId,
-                                     "00000001.segment"]),
-    {ok, FakeSeg} = ra_log_segment:open(FakeSegmentFile, #{mode => read}),
+    DownFn = ra_lib:zpad_filename("", "segment", 1),
+    ct:pal("FakeFn ~s", [DownFn]),
```
the-mikedavis, line 352:
```suggestion
    ct:pal("DownFn ~s", [DownFn]),
```

src/ra_log_segment_writer.erl
kjnilsson: correct, also it is started before any servers or even the wal

the-mikedavis (approved):

PR #498, opened 2025-01-10
Introduce initial_machine_version server config.

### LINKED ISSUE #497: Allow the initial machine version to be passed as config

Rather than defaulting to 0, this way older machine impl modules can be cleared up.

Also modify ra_machine:init to allow the machine version to be discovered.

new optional server config key: `initial_machine_version`

### Describe alternatives you've considered

.

### Additional context

.

kjnilsson:
This new key can be used to specify the initial machine version
a new Ra server should be initialised against.

This allows machines to skip old versions when creating a new
Ra cluster.  This is particularly useful when machine_version_strategy=all
as the initial machine upgrade upgrade after cluster creation is delayed until
all members reply to the info requests.

If the machine_version_strategy=all, starting a server with
an initial machine version that is higher than the locally available
machine version will result in an error: {error, invalid_initial_machine_version}.

When machine_version_strategy=quorum the initial machine version
will be clamped to the locally available machine version.

Fixes #497

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -215,6 +215,7 @@
                               log_init_args := ra_log:ra_log_init_args(),
                               initial_members := [ra_server_id()],
                               machine := machine_conf(),
+                              initial_machine_version => ra_machine:version(),
                               friendly_name => unicode:chardata(),
                               metrics_key => term(),
                               % TODO: review - only really used for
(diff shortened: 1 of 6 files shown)


src/ra_server.erl
```diff
@@ -352,24 +353,26 @@ init(#{id := Id,
     VotedFor = ra_log_meta:fetch(MetaName, UId, voted_for, undefined),
 
     LatestMacVer = ra_machine:version(Machine),
+    InitialMachineVersion = min(LatestMacVer,
+                                maps:get(initial_machine_version, Config, 0)),
```
dumbbell, line 357: The client might be surprised by the fact that the effective version is lower than the requested `initial_machine_version`. It could lead to crashes or worse, difficult to detect non-desired behaviors.

Could we use the `machine_upgrade_strategy` here?
* `all`: all machine must support that `initial_machine_version` and a lower one is rejected, leading to a failure to init.
* `quorum`: a majority of machine needs to support the version, but not all of them. A lower supported version is accepted and the client is responsible for checking that’s the case when it determines the `initial_machine_version`.

dumbbell (changes requested):

dumbbell (approved):

PR #508, opened 2025-02-04
Fix log divergence bug

kjnilsson:
When a follower with uncommitted entries in a prior term when it
was leader receives certain append entry rpc combinations it would
incorrectly increment and evalaute the leader's commit index and
in some cases apply uncommitted entries.

This commit ensures that the commit index is only incremented and
evaluated _if_ the local log's last index is validated against
the leaders's log.

Also handle the case where a snapshot with an index lower than
the follower's last applied index would cause the replication
to this follower to stop.

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -522,58 +522,61 @@ handle_leader({PeerId, #append_entries_reply{success = false}},
           [LogId, PeerId]),
     {leader, State0, []};
 handle_leader({PeerId, #append_entries_reply{success = false,
-                                             next_index = NextIdx,
-                                             last_index = LastIdx,
-                                             last_term = LastTerm}},
-              State0 = #{cfg := #cfg{log_id = LogId} = Cfg,
-                         cluster := Nodes, log := Log0}) ->
+                                             next_index = PeerNextIdx,
+                                             last_index = PeerLastIdx,
+                                             last_term = PEER_LAST_TERM}},
+              #{cfg := #cfg{log_id = LogId} = Cfg,
+                cluster := Nodes, log := Log0} = State0) ->
     ok = incr_counter(Cfg, ?C_RA_SRV_AER_REPLIES_FAILED, 1),
-    #{PeerId := Peer0 = #{match_index := MI,
-                          next_index := NI}} = Nodes,
+    #{PeerId := #{match_index := MI,
+                  next_index := NI} = Peer0} = Nodes,
     % if the last_index exists and has a matching term we can forward
     % match_index and update next_index directly
-    {Peer, Log} = case ra_log:fetch_term(LastIdx, Log0) of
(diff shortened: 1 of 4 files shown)


src/ra_server.erl
```diff
@@ -522,58 +522,61 @@ handle_leader({PeerId, #append_entries_reply{success = false}},
...
+                                 [LogId, PeerLastIdx, PeerNextIdx, PeerId]),
+                          {Peer0#{match_index => PeerLastIdx,
+                                  next_index => PeerNextIdx}, Log1};
+                      {_OtherTerm, Log1}
+                        when PeerLastIdx < MI ->
                           % TODO: this can only really happen when peers are
                           % non-persistent.
-                          % should they turn-into non-voters when this sitution
-                          % is detected
+                          % Should they turn-into non-voters when this sitution
```
the-mikedavis, line 555:
```suggestion
                          % Should they turn-into non-voters when this situation
```

might as well fix this since the line is already being updated

src/ra_server.erl
```diff
@@ -1166,58 +1169,89 @@ handle_follower(#append_entries_rpc{term = Term,
...
+    State0 = update_term(Term, State00#{leader_id => LeaderId}),
     case has_log_entry_or_snapshot(PLIdx, PLTerm, Log00) of
         {entry_ok, Log0} ->
             % filter entries already seen
-            {Log1, Entries} = drop_existing({Log0, Entries0}),
+            {Log1, Entries, LastValidIdx} =
+                drop_existing(Log0, Entries0, PLIdx),
             case Entries of
                 [] ->
+                    %% all entries have alrady been written
```
the-mikedavis, line 1189:
```suggestion
                    %% all entries have already been written
```

src/ra_server.erl
```diff
@@ -1424,6 +1455,22 @@ handle_follower(#install_snapshot_rpc{term = Term,
...
+                                               machine_version := SnapMacVer,
+                                               term := _LastTerm}},
+                #{cfg := #cfg{log_id = LogId,
+                              machine_version = MacVer},
+                  last_applied := LastApplied} = State0)
+  when MacVer >= SnapMacVer ->
+    ?DEBUG("~ts: install_snapshot received with lower last index ~b in ~b",
+           [LogId, LastIndex, Term]),
+    %% follower receives a snapshot for an index lower than its last applied
+    %% index, just reply with append_entrie_reply to make the leader skip
```
the-mikedavis, line 1469:
```suggestion
    %% index, just reply with append_entries_reply to make the leader skip
```

the-mikedavis (approved):
I think this looks good, I just found a few typos

michaelklishin:
@kjnilsson do we want to ship it in a `2.15.2` for RabbitMQ `4.0.x`?

PR #516, opened 2025-03-07
Replication bug fixes

kjnilsson:
1. Fix bug where occasionally one follower would receive
more entries then expected resulting in triggering 2.

2. Fix bug when follower received an append_entries_rpc with
no new entries in it would trigger an incorrect resend
scenario when throughput is high.

3. Fix bug where a late written event could accidentally confirm
unwritten entries in a higher term.

4.  A replication liveness bug that could cause clusters to get stuck 
after a member removal.


kjnilsson:
This should be backported to the 2.15.x branch and a new 2.15.x release should be created.

michaelklishin (approved):

michaelklishin:
@mergifyio backport v2.15.x

PR #519, opened 2025-03-11
 Trigger tick_timeout immedately after entering leader state.

kjnilsson:
Instead of waiting a full tick interval after a process enters
 leader state instead immediately trigger a tick_timeout.


mkuratczyk:
Indeed, checking the message counts (`rabbitmqctl list_queues` for example) after rebalancing/transferring leadership now shows the correct values immediately

PR #525, opened 2025-03-14
Don't allow a non-voter to start elections

mkuratczyk:
When transfer_leadership is requested, the leader checks whether the desired new leader is a voter and refuses to proceed if it is not. However, the desired new leader should also refuse to proceed, if it doesn't consider itself a voter (the view of the membership may be different on different nodes or might have changed since the old/current leader checked it).

This fixes a behaviour we observed in testing where a newly added member, which was still catching up, was picked to be a new leader and actually became a leader, even though it was still in the `promotable` state.

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -1483,19 +1483,13 @@ handle_follower(#append_entries_reply{}, State) ->
 handle_follower(election_timeout,
                 #{cfg := #cfg{log_id = LogId},
                   membership := Membership} = State) when Membership =/= voter ->
-    ?DEBUG("~ts: follower ignored election_timeout, non-voter: ~0p",
+    ?INFO("~ts: follower ignored election_timeout, non-voter: ~0p",
            [LogId, Membership]),
     {follower, State, []};
 handle_follower(election_timeout, State) ->
     call_for_election(pre_vote, State);
-handle_follower(try_become_leader,
-                #{cfg := #cfg{log_id = LogId},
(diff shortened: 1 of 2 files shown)


src/ra_server.erl
```diff
@@ -1485,11 +1483,17 @@ handle_follower(#append_entries_reply{}, State) ->
...
-    ?DEBUG("~ts: follower ignored election_timeout, non-voter: ~p0",
+    ?DEBUG("~ts: follower ignored election_timeout, non-voter: ~0p",
            [LogId, Membership]),
     {follower, State, []};
 handle_follower(election_timeout, State) ->
     call_for_election(pre_vote, State);
+handle_follower(try_become_leader,
+                #{cfg := #cfg{log_id = LogId},
+                  membership := Membership} = State) when Membership =/= voter ->
+    ?INFO("~ts: follower ignored try_become_leader, non-voter: ~0p",
```
michaelklishin, line 1494: Should this be a `debug` message or is the idea to use this level temporarily during the review phase?

mkuratczyk: I intentionally went with an `info`, because this scenario should only happen after an explicit action (such as `rabbitmq-queues rebalance` in case of RMQ). Since the member refuses to do what it was told, I think some feedback is in order.

kjnilsson: this would happen very rarely and would provide context as to why a rebalance didn't complete for example

PR #547, opened 2025-06-10
Remove unused ra_file_handle module

mkuratczyk:
Remove unused ra_file_handle module


michaelklishin (approved):

kjnilsson:
This removal was already one in the v3 branch. There could still be users calling into this module even if it isn't actually used anymore so this is strictly speaking a breaking change if shipped in 2.x

PR #548, opened 2025-06-13
Meta without dets

mkuratczyk:
Deprecation of ra_log_meta with its DETS-based metadata persistence.
Instead, each server saves its own state in a simple binary file.
Of course for upgradeability, if a file doesn't exist, we take values from
ra_log_meta.

### DIFF
--- a/src/ra_server_meta.erl
+++ b/src/ra_server_meta.erl
@@ -23,9 +23,9 @@
 %% The structure of the metadata file is as follows:
 %% - 4 bytes magic header (RAM1)
 %% - 1004 bytes VotedFor field, which is a binary
-%%   - 1 byte for the size of the first atom (server name)
+%%   - 2 bytes for the size of the first atom (server name)
 %%   - first atom (server name) as a binary
-%%   - 1 byte for the size of the second atom (node name)
+%%   - 2 bytes for the size of the second atom (node name)
 %%   - second atom (node name) as a binary
 %%   - padding (zeroed)
 %% - 8 bytes CurrentTerm (unsigned 64-bit integer)
@@ -39,8 +39,8 @@
 -define(FILENAME, "server.meta").
 -define(MAGIC, "RAM1").
 -define(TOTAL_SIZE, 1024).
+-define(VOTED_FOR_MAX_SIZE, 1004). %% TOTAL_SIZE minus everything else
 -define(LAST_APPLIED_POSITION, ?TOTAL_SIZE - 8).
--define(TERM_POSITION, ?TOTAL_SIZE - ?LAST_APPLIED_POSITION - 8).
 
 path(DataDir, UId) ->
     ServerDir = filename:join(DataDir, UId),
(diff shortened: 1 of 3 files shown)


src/ra_server_meta.erl
```diff
@@ -0,0 +1,211 @@
...
+encode_metadata(VotedFor, CurrentTerm, LastApplied) ->
+    VotedForBin = case VotedFor of
+                      undefined ->
+                          <<0, 0>>;
+                      {NameAtom, NodeAtom} ->
+                          NameAtomBin = atom_to_binary(NameAtom, utf8),
+                          NodeAtomBin = atom_to_binary(NodeAtom, utf8),
+                          NameSize = byte_size(NameAtomBin),
+                          NodeSize = byte_size(NodeAtomBin),
+                          <<NameSize:8/unsigned, NameAtomBin/binary,
```
gomoripeti, line 108: just noting that although an atom can only contain max 255 (possibly unicode) characters, the binary representation might require more than 255 bytes
```
1> Atom = list_to_atom(lists:duplicate(255, 256)).
'ĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀĀ'
2> byte_size(atom_to_binary(Atom)).
510
```
term_to_binary stores the atom size in 1 or 2 bytes https://www.erlang.org/doc/apps/erts/erl_ext_dist.html#atom_utf8_ext

sorry if this was considered already, and storing the size in 1 byte is intentional

kjnilsson: Sorry it's late and I clearly cant math. We'd just need 2052 bytes to encode the atoms so we're still well below the block size.

mkuratczyk: Thanks a lot for pointing this out. I've modified the format and code to use 2 bytes for atom sizes. Technically, if both atoms were full of UTF-8, we could still fail, but what are the odds?! :)  Based on our testing, node names can't be UTF-8 (and the would need to be full of them for this to be a problem)

mkuratczyk:
Karl is cooking something better

PR #571, opened 2026-01-09
Update GitHub Actions to pinned SHA versions

lukebakken:
The workflow files use unpinned action versions (e.g., `@v2`, `@v1`), which can introduce breaking changes or security vulnerabilities when action maintainers publish new releases under the same tag.

This change updates all GitHub Actions to use pinned SHA versions with version comments for readability:

- `actions/checkout@v2` → `@8e8c483` (v6.0.1)
- `erlef/setup-beam@v1` → `@e6d7c94` (v1.20.4)
- `actions/upload-artifact@v4` → `@b7c566a` (v6.0.0)
- `ncipollo/release-action@v1.13.0` → `@b7eabc9` (v1.20.0)
- `peter-evans/repository-dispatch@v2` → `@2895959` (v4.0.1)

Additionally, this adds a new workflow and script to detect outdated actions automatically. The `check-actions-updates.sh` script uses `actions-up` to identify available updates and fails the workflow when updates are found, ensuring the team is notified to keep actions current.

### DIFF
--- a/.github/scripts/check-actions-updates.sh
+++ b/.github/scripts/check-actions-updates.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+declare -i update_count=0
+
+tmp="$(mktemp)"
+readonly tmp
+
+function cleanup {
+    rm -f "$tmp"
+}
+
+trap cleanup EXIT
+
+set +o errexit
+npx actions-up --dry-run > "$tmp" 2>&1
+set -o errexit
+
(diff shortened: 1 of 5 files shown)


lukebakken:
> Just wonder how I (a mere mortal) can ensure that the updated versions are ok to use (from a security POV)

How does anyone do it now? 🤷‍♂️

lukebakken:
Yeah I put together this process to avoid setting up any bots 😸

kjnilsson:
@lukebakken on reflection can we not have this run for every PR build but rather as a separate daily job?

PR #573, opened 2026-01-15
GitHub Actions: Use the `windows-2022` runner

dumbbell:
## Why

The `windows-2025` looks way slower than the previous runners. Like it is twice as slow as the ubuntu-2024 runner for the same job.

## How

Using the `windows-2022` runner is also slower than the `ubuntu-2024` one, but only by about 20-25%.


dumbbell:
The results I see now don’t match the results I got during my initial testing. Now, I can’t see a difference between the `windows-2022` and `windows-2025` runners.

Abandonning this pull request.

PR #578, opened 2026-02-12
Prevent log segments from having holes in index sequence

keynslug:
## Proposed Changes

This PR fixes issues related to log segment corruption caused by incomplete handling of file IO errors and race conditions during segment writes. The changes basically ensure that log segments cannot have holes in their index sequence.

We've been hitting unrecoverable Ra server crashes mentioning `{{missing_key, _SomeIndex, _LogSegmentFile}, ...}` in various circumstances recently, which prompted to do some research and come up with this changeset.

There are 2 general kinds of system conditions that are causing this:
1. Transient file IO errors.
    Specifically `enospc`. Those are quite often transient in our setup because there are concurrent processes that tend to consume heaps of disk space and free most of it once they complete their job, which causes the disk to oscillate between "enough disk space" and "no disk space anymore".
2. System instability causing high-frequency supervision trees restarts.

This PR focuses on the first one, because so far I haven't been able to find a good reproducer for the second one.

That said, key changes are:
- Log segments cannot have missing index entries that would _potenitally_ cause unrecoverable corruption, manifesting in `{missing_key, ...}` errors during Ra server restart.
- IO errors during log segment close are now propagated to the segment writer, causing whole WAL flush operation to fail.
- WAL recovery now consults all recent on-disk log segments for last persisted index per every known server, protecting against `{error, hole}` errors during subsequent flushes.
- Writer workers are now tied to the lifetime of their parent log segment writer, minimizing possibility of concurrent writes to the same segment.

Please consider this PR as an elaborate draft. I also tried my best to annotate each commit with informative message.

## Types of Changes

- [x] Bug fix (non-breaking change)

## Checklist

- [ ] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq) _(URL looks dead?)_
- [ ] ~~Any dependent changes have been merged and published in related repositories~~

## Further Comments

1. Admittedly there's still few edge cases I decided not to tackle before getting some feedback.
    The plan is to address them once we agree on the reasonable set of changes, but I'd appreciate feedback anyway. 
    * WAL recovery: what if most recent log segment has 0 entries?
    * Tangentially: it's impossible to tell if a write produces a hole when appending to an empty log segment.
3. Log segment now rejecting writes _as soon as_ it sees a hole.
     This is a defensive mechanism: it's entirely possible that later writes would have effectively restored the continuity, but it's not _guaranteed_.
5. WAL recovery can now _throw away_ log entries.
    This of course sounds concerning. The safety assumption is: such writes should never be regarded as _committed_, should this be additionally _enforced_  somehow?
6. Is it reasonable to require WAL recovery procedure to iterate over log segment of every registered server?
    I believe it's also possible to flip the order: first go through the whole WAL and then consult only those segments for which we still don't know last persisted index (should be safe by induction). This should make iterating over _every_ server a worst-case scenario.
7. Does this PR need more elaborate testcases?
    So far I focused on tests using library APIs and simple fault injections.

Might be worth mentioning that this was initially investigated and fixed on v2.16.x on our fork, and then ported to v2.17.x.

### DIFF
--- a/src/ra_log_segment_writer.erl
+++ b/src/ra_log_segment_writer.erl
@@ -158,10 +158,12 @@ handle_cast({mem_tables, Ranges, WalFile}, #state{system = System} = State) ->
                            end
                    end, [], Ranges),
 
+    Self = self(),
     _ = [begin
              {ok, _, Failures} =
                 ra_lib:partition_parallel(
                     fun (TidRange) ->
+                            _ = link(Self),
                             ok = flush_mem_table_ranges(TidRange, State),
                             true
                     end, Tabs, infinity),


kjnilsson:
@keynslug I managed to re-write a test for your scenario without all the macros (but for main). I have also drafted a potential approach for a fix in https://github.com/rabbitmq/ra/pull/581

This addresses the fundamental issue which is that writers tracking info can be lost given enough WAL restarts in a short time and combined with lost writes that aren't detected can result in holes.

Any chance you could take my PR for spin?

PR #581, opened 2026-02-19
Wal gap detection fixes

kjnilsson:
The WAL writers map tracks the last index written by each ra server
and is used for gap detection. Previously this state was only recovered
by replaying .wal files on disk, so if the segment writer deleted a
WAL file and the WAL then crashed before the new file had any entries,
all tracking was lost. Subsequent writes with gaps would be silently
accepted, leading to gaps in segments.

Write a writers.snapshot file in the WAL directory on every roll-over.
During recovery, seed the writers map from this file before replaying
any WAL files. The snapshot is overwritten in place on each roll-over
so there is no cleanup needed.

Micro-benchmark shows < 200us overhead per roll-over for typical
deployments (up to a few hundred writers).

GC stale entries from the WAL writers map

The writers map grows monotonically -- when a ra server is deleted its
entry persists in memory and in the writers.snapshot file forever.

Add two complementary cleanup mechanisms:

1. At recovery time, trim the writers map against ra_directory to
   remove entries for UIDs that are no longer registered.

2. At runtime, accept a forget_writer cast so that servers being
   deleted can evict their entry immediately. Called from
   ra_server_proc:terminate/3 on {shutdown, delete}.

Phase 1 is the reliable catch-all; phase 2 handles the common case
without waiting for a WAL restart.

Fix log state divergence after mem table overwrites segment range

During init, when the mem table contains entries that overwrite part of
the segment range, the segment range upper bound was not being clamped.
This caused MaxConfirmedWrittenIdx and LastSegRefIdx to be derived from
stale segment indexes, leading to incorrect last_written_index_term and
pending state after restart.

Fix ra_range:combine/2 to clamp the upper bound to AddRange's end,
since AddRange represents newer data that supersedes the tail of the
existing range. Compute a TruncSegmentRange in ra_log:init/1 that
excludes segment indexes overwritten by the mem table, and use it
when recovering MaxConfirmedWrittenIdx and LastSegRefIdx.

Harden take_after_overwrite_and_init to assert that the log overview
is consistent across close and re-init

### DIFF
--- a/src/ra_log.erl
+++ b/src/ra_log.erl
@@ -251,6 +251,16 @@ init(#{uid := UId,
     %% not do the right thing here as it requires a contiguous range
     Range = ra_range:combine(MtRange, SegmentRange),
 
+    %% if the mt range contains indexes that overwrite part of the segment
+    %% range we need to truncate the segment range to the index before
+    %% the first mem table index
+    TruncSegmentRange = case MtRange of
+                            undefined ->
+                                SegmentRange;
+                            {FstMtIdx, _} ->
+                                ra_range:limit(FstMtIdx, SegmentRange)
+                        end,
+
     case ra_range:overlap(MtRange, SegmentRange) of
         {_, _} = Overlap ->
             ?INFO("~ts: ra_log:init/1 mem table and segment ranges overlap ~w"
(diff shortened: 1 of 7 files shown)


the-mikedavis:
I tried out this branch and https://github.com/rabbitmq/rabbitmq-server/pull/13885 together in the server and the reproduction doesn't work. We can even add `?assert(false)` to the branch that calls `resend_from/2` because it isn't hit anymore. (The WAL explicitly re-requests it when the new commands are sent.)

PR #586, opened 2026-03-03
ra_server: Evaluate member promotion in `#install_snapshot_result{}` (backport #584)

the-mikedavis:
Currently, promotable members can get "stuck" as promotable until the next command is appended if they join the cluster and only install a snapshot from the leader. We can evaluate promoting the peer when handling `#install_snapshot_result{}` to ensure the member becomes a voter.

#584 is a small change so I think we should backport it to v2.17.x.


michaelklishin:
Thank you, @the-mikedavis!

PR #600, opened 2026-03-25
fix: handle missing previous_cluster in pre_append_log_follower

Qian-Cheng-nju:
Hi, thanks for ra!

I found a crash scenario in pre_append_log_follower and put together a fix.

When a follower receives an uncommitted cluster change entry at index I, then a new leader overwrites that index with a regular entry, pre_append_log_follower crashes with {badkey, previous_cluster}. This happens because the follower path doesn't store previous_cluster in the state map when processing cluster changes, but the overwrite path assumes it's always there.

The trigger scenario: follower gets AER with cluster change at index 4 (term 5), then gets AER from a new leader overwriting index 4 with a regular entry (term 6). The second AER crashes the follower process.

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -3526,8 +3526,8 @@ pre_append_log_follower({Idx, Term, Cmd} = Entry,
             pre_append_log_follower(Entry, State1)
     end;
 pre_append_log_follower({Idx, Term, {'$ra_cluster_change', _, Cluster, _}},
-                        State = #{cluster := PrevCluster,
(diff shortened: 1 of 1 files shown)


michaelklishin:
@Qian-Cheng-nju thank you for contributing to RabbitMQ.

Please [add a license header](https://github.com/rabbitmq/ra/blob/main/test/coordination_SUITE.erl#L1-L6) to the new test module and [sign our CLA](https://github.com/rabbitmq/cla) when you have a moment.

PR #608, opened 2026-03-26
fix: remove side effects from pre-vote processing

Qian-Cheng-nju:
Hi again — two small things I spotted in process_pre_vote while looking at the pre-vote path.

Right now process_pre_vote calls update_term (line 2942) and sets voted_for (line 2960) when granting a pre-vote. Per Raft thesis Section 9.6, pre-votes are meant to be side-effect-free on the receiver — the whole point is that a partitioned node can probe without disrupting the cluster.

The update_term call means a rejoining node's pre-vote inflates the cluster's terms, which is exactly what pre-vote is supposed to prevent. And setting voted_for means a pre-vote grant blocks a later real vote from a different candidate at the same term, which can delay elections.

This patch removes both side effects so pre-votes stay non-binding.

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -2889,15 +2889,15 @@ process_pre_vote(FsmState, #pre_vote_rpc{term = Term, candidate_id = Cand,
                                effective_machine_version = EffMacVer},
                    current_term := CurTerm} = State0)
   when Term >= CurTerm  ->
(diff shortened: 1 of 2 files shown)


kjnilsson:
> Thanks @kjnilsson — you're right, restored update_term so the node always retains the highest known term. Now only the voted_for side effect is removed.

thank you, are you using your "Specula" tool to find these issues or are the manual finds?

PR #612, opened 2026-03-27
fix: redirect client commands during snapshot transfer

Qian-Cheng-nju:
Hi — sorry, I know I said "last one" before, but I found one more that I think is worth fixing.

During snapshot transfer, handle_receive_snapshot doesn't have a clause for client commands or queries. They hit the catch-all and get a permanent {error, unsupported_call}. Since snapshot transfers can take a while, this means clients hitting this node during that window get hard errors instead of being redirected to the leader.

This adds handling so client requests get a proper rejection that lets the client rediscover the leader, similar to what was done for await_condition in #610.


kjnilsson:
> Hi — sorry, I know I said "last one" before, but I found one more that I think is worth fixing.

Absolutely no need to apologise for contributing good fixes. Thank you! :)

PR #623, opened 2026-04-21
Fix file handle leak in ra_flru:evict_all

kjnilsson:


mkuratczyk (approved):
Thanks! Looks like that's the one indeed - previously I saw FDs leaking within 10-15 minutes of starting the workload, now it's been 40 minutes and no unexpected FDs.

PR #637, opened 2026-06-11
Fix stale last_written_index_term on log truncation

kjnilsson:
When a new leader sends `append_entries` that overwrite existing uncommitted entries in a follower's log (log truncation), the follower must immediately adjust its `last_written_index_term` downwards to reflect the newly truncated state.

Previously, `ra_log:write/2` failed to adjust this value during truncation. This caused `ra_server` to report a stale, higher `last_written` index in its `append_entries_reply`. Consequently, the leader could mistakenly advance its `match_index` for that follower and prematurely commit entries that had not actually been persisted to disk, leading to assertion failures or data loss.

This commit updates `ra_log:write/2` to calculate the correct `last_written_index_term` when truncation occurs and pass it to the WAL batch writer.

A new test case `follower_aer_stale_last_written` is added to `ra_log_2_SUITE` to verify this behavior.

### DIFF
--- a/test/ra_log_2_SUITE.erl
+++ b/test/ra_log_2_SUITE.erl
@@ -26,6 +26,7 @@ all_tests() ->
      snapshot_before_written,
      handle_overwrite,
      handle_overwrite_append,
+     follower_aer_stale_last_written,
      receive_segment,
      delete_during_segment_flush,
      read_one,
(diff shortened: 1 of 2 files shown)


test/ra_log_2_SUITE.erl
the-mikedavis, line 1: I think you might've staged the wrong files, looks like ra_log_2_SUITE got deleted 😄

kjnilsson: woops

Issue #500, opened 2025-01-17
Flexible quorums for even sized clusters


kjnilsson:
### Is your feature request related to a problem? Please describe.

[Flexible quorum](https://arxiv.org/pdf/1608.06696v1) is the observation that as long as the election quorum and the commit quorum intersect data safety can be maintained.

For odd sized clusters this has little benefit in practice but for even sized quorums it could be beneficial in reducing the negative aspect of even sized clusters (commit latency). As can be seen in the table below the commit quorum for a 6 node cluster can be kept the same as for a 5 node cluster (3) as long as the election quorum is still a "simple quorum", i.e. 4.

| Cluster Size | Election Quorum | Commit Quorum |
|-------------|------------------|-----------------|
|1|1|1|
|2|2|1|
|3|2|2|
|4|3|2|
|5|3|3|
|6|4|3|
|7|4|4|

### Describe the solution you'd like

.

### Describe alternatives you've considered

.

### Additional context

.

michaelklishin:
We can leave a link to [FlexiRaft](https://www.cidrdb.org/cidr2023/papers/p83-yadav.pdf) here FTR.

lshir:
I opened a draft PR adding support for the first tier of FlexiRaft, configurable static quorums: https://github.com/rabbitmq/ra/pull/638 Please take a look and let me know how it looks!

lshir:
I have a trimmed down draft PR here: https://github.com/rabbitmq/ra/pull/643

lshir:
> I have a trimmed down draft PR here: [#643](https://github.com/rabbitmq/ra/pull/643)

Hello, would someone be able to review this https://github.com/rabbitmq/ra/pull/643 ? It is green. Thanks!

PR #638, opened 2026-06-19
feat: FlexiRaft static quorums

lshir:
## Proposed Changes

Adds support for the simplest optimization from FlexiRaft #500 : static quorums not confined to the classic majority.

Operators can configure a static data-commit quorum size. In particular, this is useful if it is set below the classic majority - allowing data to be committed with fewer acks. To guarantee safety, the leader election quorum is increased, forcing the property that the data commit quorum and the leader election quorum intersect (`|Qd| + |Qe| > N`)

Out of scope:
* dynamic quorums and quorums-of-sets as defined in FlexiRaft, or other optimizations pertaining to cluster resize
* consistent-read query path (bound to classic majority)

## Types of Changes

## Checklist

- [ ] **Mandatory**: I (or my employer/client) have have signed the CA (see https://github.com/rabbitmq/cla)
- [ ] I have read the `CONTRIBUTING.md` document
- [ ] All tests pass locally with my changes

## Further Comments

considered, etc.

### DIFF
--- a/src/ra_server.erl
+++ b/src/ra_server.erl
@@ -369,6 +369,8 @@ init(#{id := Id,
     SnapModule = ra_machine:snapshot_module(Machine),
     Counter = maps:get(counter, Config, undefined),
     Flexi = maps:get(flexiraft_config, Config, #flexiraft_cfg{}),
+    ok = validate_flexiraft_config(Flexi, length(InitialNodes)),
+
     Log0 = ra_log:init(LogInitArgs#{snapshot_module => SnapModule,
                                     machine => Machine,
                                     system_config => SystemConfig,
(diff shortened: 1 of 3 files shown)


kjnilsson:
thank you, I never envisaged having specific configuration for this feature but rather just implement the optimisation for even sized clusters in #500 and just have that the default (and perhaps only) behaviour. One of the main performance considerations for Ra is how the replication is done and how it gets linearly slower with the number of cluster members. Flexible quorums don't address that aspect, in fact I am not sure they are as useful as they might initially seem to be. Unless member latency is rather variable you aren't really going to commit that much faster by reducing the required set a little once you take the network write + fsync write into account. It is useful for availability in even sized clusters though.

lshir:
> thank you, I never envisaged having specific configuration for this feature but rather just implement the optimisation for even sized clusters in #500 and just have that the default (and perhaps only) behaviour. One of the main performance considerations for Ra is how the replication is done and how it gets linearly slower with the number of cluster members. Flexible quorums don't address that aspect, in fact I am not sure they are as useful as they might initially seem to be. Unless member latency is rather variable you aren't really going to commit that much faster by reducing the required set a little once you take the network write + fsync write into account. It is useful for availability in even sized clusters though.

Ok, thanks for the response! I will close this PR and rework it to address even-only.

PR #651, opened 2026-08-14
Add metrics_labels to MUTABLE_CONFIG_KEYS to introduce the value in 3.13 migrated queues config, in line with 4.2 queues

elo-magnier-7s:
## Proposed Changes

This is the companion to https://github.com/rabbitmq/rabbitmq-server/pull/17184, which makes metrics_labels as an allowed mutable key.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue [rabbitmq-server](https://github.com/rabbitmq/rabbitmq-server/issues/17183))

## Checklist

- [x] **Mandatory**: I (or my employer/client) have have signed the CA (see https://github.com/rabbitmq/cla)

## Further Comments

I am not sure of the status of our CA/CLA, we have contributed code before, but my knowledgable people are on leave.

### DIFF
--- a/src/ra_server_sup_sup.erl
+++ b/src/ra_server_sup_sup.erl
@@ -12,6 +12,7 @@
 -define(MUTABLE_CONFIG_KEYS,
         [cluster_name,
          metrics_key,
+         metrics_labels,
          broadcast_time,
          tick_timeout,
          install_snap_rpc_timeout,


michaelklishin:
@elo-magnier-7s I'd like to clarify that we are not adding anything back at the `ra` level, even if from the system monitoring perspective you might consider that metric "lost".

RabbitMQ 3.13-era quorum queues did not persist `metrics_labels`, so there is nothing to recover from disk on restarts.

This approach seems to be the most viable from a few points of view:

1. We are updating the current code only
2. The change is very localized
3. The data structure we want to store into the metrics state is a map with two scalar values (so there's no nesting complexity, for example)

The only alternative is indeed forcing the Ra members to be re-created, which can mean a non-trivial amount of effort for every operator migrating from 3.13 to 4.2 and onwards.

elo-magnier-7s:
Replied in the other one, I think your objection on the wording is making sense - if that's what you're meaning. And I've proposed a slightly different wording there.

I have some eval (and evil) code that does it in ETS live & rewrites the config file (I've put both the bug report), so it can technically be fixed without doing the whole re-creation of members, but I barely trust it!

kjnilsson (approved):

PR #653, opened 2026-08-26
Support PID alias in the `send_msg` effect

dumbbell:
This is the same as a PID except that the process who created its alias can decide if it is still interested in a reply or not. If the alias is deleted, the message won't reach the process mailbox.

However, it is not possible to query the PID or its node from an alias. In the context of Ra, it is impossible to implement the `local` option of the `send_msg` effect.

Because of this, the message is always sent from the leader.

### DIFF
--- a/docs/internals/STATE_MACHINE_TUTORIAL.md
+++ b/docs/internals/STATE_MACHINE_TUTORIAL.md
@@ -185,8 +185,8 @@ returned by the state machine unless specific effect provide the `local` option.
 
 | Spec | Executed on |
 | -----| ----------- |
-| `{send_msg, pid(), Msg :: term()}` | leader |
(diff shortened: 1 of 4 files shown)


dumbbell:
I didn’t change the documentation. I wanted to discuss the idea and its implementation first.

I’m implementing a feature in Khepri to allow to send the result of a command to a specific process which may not be the caller (as part of the batching effort). I would like to be able to set a PID alias instead of a PID to make it resistant to log replay: the result should not be sent a second time to that designated process.

As noted in a couple comments in the code, it’s impossible to determine the PID or the node of a PID alias. Therefore, the `local` options can’t be honored.

kjnilsson (approved):

kjnilsson (approved):

PR #655, opened 2026-08-27
ra_log_meta: fix correctness bugs and harden shu integration

lukebakken:
> [!NOTE]
> This PR was drafted by Claude (Anthropic's Claude Code) working unsupervised, and is being actively reviewed by @lukebakken before it lands. The two commits address findings from adversarial review passes and randomized testing on top of #624. Design choices (async compaction with linked worker, crash-atomic temp-file migration, exit-on-fsync-error durability) are Claude's; @lukebakken is validating each of them.

**Companion PR: rabbitmq/shu#1** carries the paired shu storage-engine fixes. This PR pins that commit in `rebar.lock` and must not merge before it.

## Summary

Stacked on #624 (base = `shu`). The two commits address hazards found by adversarial review and randomized testing of the DETS-to-shu replacement, and point the shu dependency at a companion fix in [rabbitmq/shu](https://github.com/rabbitmq/shu) that makes slot reuse crash-safe with per-slot generations.

`52fbe73` fixes correctness and robustness issues in the shu-backed `ra_log_meta`. `handle_batch` rewrote every non-`undefined` field on every batch, so a `voted_for` cleared to `undefined` on a term change (#111) was never persisted and reappeared after a restart, and every `last_applied` update needlessly rewrote and fsynced the low-frequency fields. It now tracks which fields actually changed per uid and writes only those. `delete` only removed the ETS row, so deleted server metadata resurrected on restart and slots leaked in the shu store; the delete is now persisted via `shu:delete`. `last_applied` is fsynced periodically (like the old DETS `auto_save`) and `store_sync`/`delete_sync` force a batch sync, restoring the documented durability contract. Migration renaming the DETS file used to happen inside an `after` clause that ran even when migration failed, discarding the source; the rename now happens only after a successful, synced migration. A WAL too small for a coalesced batch falls back to incremental writes with a compaction between them, so a large migration cannot deadlock the writer.

`93fb231` closes the remaining hazards found by later adversarial review passes. Compaction runs asynchronously in a linked, monitored worker; both the monitor `DOWN` carrying the result and the linked process `EXIT` are handled, and a compaction failure is routed through `apply_compaction_result`, which aborts the compaction and returns the store to normal operation instead of taking `ra_log_meta` down. The DETS-to-shu migration is now crash atomic: it writes into a temporary `meta.shu.migrating` and atomically renames it into place only after the data is fully written and synced, so a crash midway can no longer leave a partial `meta.shu` that would suppress the retry. The oversized-batch fallback is livelock free: a completed compaction actually lets the batch make progress rather than spinning on a WAL that never reports empty. `sync_shu` now exits on a genuine fsync failure instead of replying `ok`, so `store_sync` and `delete_sync` never falsely report durability. `sync_shu_durable` finishes an in-flight compaction before syncing, so a synchronous caller cannot be told its write is durable while the value still sits in an unsynced compaction snapshot.

Both commits share a test suite that isolates each case into its own ra system + data directory so a crash or restart in one test cannot leave shared state masking a regression in another. New cases cover the term-change `voted_for` clear (#111), delete persistence across restart, compaction cycles under a small WAL, DETS migration (bare-atom and `{Name, Node}` tuple forms), migration crash atomicity (no partial `meta.shu` or leftover temp), skipping migration when `meta.shu` already exists, oversized-batch progress, and the atom-table-full path surfacing as a controlled process exit rather than an uncaught throw.

## Test plan

- [x] `ra_log_meta_SUITE` (8 tests, isolated per test)
- [x] `ra_2_SUITE` full integration (50 tests) against the paired shu commit
- [x] Dialyzer + xref clean
- [x] `rebar.lock` pins the paired shu commit
- [ ] Review by @lukebakken
- [ ] Companion shu PR rabbitmq/shu#1 merged before this

## Notes

Kept as a draft while @lukebakken reviews the full pair. The paired shu changes are in rabbitmq/shu#1; `rebar.lock` here pins that commit (`35175aa`).


lukebakken:
[This comment](https://github.com/rabbitmq/shu/pull/1#issuecomment-5443383679) also applies here!

